PR #10, opened 2017-05-24
runtime: test and fixes to run basic drivers

smola:
This PR includes fixes and tests required to run drivers:

* Tests run a privileged container.
* Image metadata is imported and later used to restore environment variables in containers.
* runtime.Bootstrap() needs to run in tests.
* Avoid requiring capabilities when we don't need to.

### DIFF
--- a/runtime/container.go
+++ b/runtime/container.go
@@ -3,6 +3,7 @@ package runtime
 import (
 	"os"
 
+	"github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/opencontainers/runc/libcontainer"
 )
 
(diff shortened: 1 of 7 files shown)


utils/config.go
```diff
@@ -0,0 +1,49 @@
...
+
+import (
+	"encoding/json"
+	"os"
+
+	"github.com/containers/image/types"
+	"github.com/opencontainers/image-spec/specs-go/v1"
+)
+
+func WriteImageConfig(img types.Image, path string) error {
```
mcuadros: this util function should be tested isolate in this package

runtime/container.go
```diff
@@ -37,20 +38,26 @@ type Command interface {
 	Wait() error
 }
 
-func newContainer(c libcontainer.Container, p *Process) Container {
+func newContainer(c libcontainer.Container, p *Process, image *v1.Image) Container {
```
mcuadros: `image` is not a Image, is just metainformation should be should change the name to something more descriptive

runtime/container.go
```diff
@@ -37,20 +38,26 @@ type Command interface {
...
 		Container: c,
 		process:   &cp,
+		image:     image,
 	}
 }
 
 type container struct {
 	libcontainer.Container
 	process *libcontainer.Process
+	image   *v1.Image
```
mcuadros: same here

PR #11, opened 2017-05-25
runtime: ensure containers get streams closed on exit

smola:


smola:
Depends on https://github.com/bblfsh/server/pull/10

runtime/container.go
```diff
@@ -37,35 +39,57 @@ type Command interface {
...
+	defer func() {
+		derr = c.Container.Destroy()
+		streams := []interface{}{
+			c.process.Stdin,
+			c.process.Stdout,
+			c.process.Stderr,
+		}
+		for _, s := range streams {
+			if c, ok := s.(io.Closer); ok {
+				if err := c.Close(); err != nil && derr != nil {
```
abeaumont: I've not enough background to judge this code, but it looks weird this `derr != nil` check, why is it checked?

smola: If there are multiple errors, only the first one is returned. So it does not override the error if it is already set.

abeaumont: then it should be `derr == nil`, otherwise you're overriding it, aren't you?

PR #14, opened 2017-05-29
Driver pool, closes #3

### LINKED ISSUE #3: add driver client pool
Create driver client pool, controlling the execution of a pool of drivers with the same image and using it to serve concurrent requests.

smola:
Depends on https://github.com/bblfsh/server/pull/12

### DIFF
--- a/pool.go
+++ b/pool.go
@@ -1,0 +1,192 @@
+package server
+
+import (
+	"sync"
+	"time"
+
+	"github.com/Sirupsen/logrus"
+	"github.com/bblfsh/sdk/protocol"
+	"github.com/pkg/errors"
+	"golang.org/x/time/rate"
+)
+
+const (
+	DefaultPoolMin               = 1
+	DefaultPoolMax               = 5
+	DefaultPoolTimeBeforeNew     = time.Second * 3
+	DefaultPoolTimeBetweenSpawns = time.Second * 10
+	DefaultPoolTimeBeforeClean   = time.Minute * 1
+)
+
+// DriverPool controls a pool of drivers and balances requests among them,
+// ensuring each driver does not get concurrent requests.
+type DriverPool struct {
+	// Min
+	Min               int
+	Max               int
+	TimeBeforeNew     time.Duration
+	TimeBetweenSpawns time.Duration
+	TimeBeforeClean   time.Duration
+	New               func() (Driver, error)
+
+	cur     int
+	ch      chan Driver
+	m       *sync.Mutex
+	close   chan struct{}
+	closed  bool
+	limiter *rate.Limiter
+	last    time.Time
+}
+
+func NewDriverPool(new func() (Driver, error)) *DriverPool {
+	return &DriverPool{
+		Min:               DefaultPoolMin,
+		Max:               DefaultPoolMax,
+		TimeBeforeNew:     DefaultPoolTimeBeforeNew,
+		TimeBetweenSpawns: DefaultPoolTimeBetweenSpawns,
+		TimeBeforeClean:   DefaultPoolTimeBeforeClean,
+		New:               new,
+	}
+}
+
+// Start starts the DriverPool and runs the minimum number of instances.
+func (dp *DriverPool) Start() error {
+	dp.ch = make(chan Driver)
+	dp.m = &sync.Mutex{}
+	dp.close = make(chan struct{})
+	dp.limiter = rate.NewLimiter(rate.Every(dp.TimeBetweenSpawns), 1)
+	dp.last = time.Now()
+	go dp.gc()
+	return dp.SetInstanceCount(dp.Min)
+}
+
+// SetInstanceCount changes the number of running driver instances. Instances
+// will be started or stopped as necessary to satisfy the new instance count.
+// It blocks until the all required instances are started or stopped.
+func (dp *DriverPool) SetInstanceCount(target int) error {
+	dp.m.Lock()
+	defer dp.m.Unlock()
+
+	n := target - dp.cur
+	if n > 0 {
+		return dp.add(n)
+	} else if n < 0 {
+		return dp.del(-n)
+	}
+
+	return nil
+}
+
+func (dp *DriverPool) add(n int) error {
+	for i := 0; i < n; i++ {
+		d, err := dp.New()
+		if err != nil {
+			return err
+		}
+
+		dp.enqueue(d)
+		dp.cur++
+	}
+
+	return nil
+}
+
+func (dp *DriverPool) del(n int) error {
+	for i := 0; i < n; i++ {
+		d := <-dp.ch
+		dp.cur--
+		if err := d.Close(); err != nil {
+			return err
+		}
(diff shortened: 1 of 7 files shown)


pool.go
```diff
@@ -0,0 +1,192 @@
...
+	// Min
+	Min               int
+	Max               int
+	TimeBeforeNew     time.Duration
+	TimeBetweenSpawns time.Duration
+	TimeBeforeClean   time.Duration
+	New               func() (Driver, error)
+
+	cur     int
+	ch      chan Driver
```
alcortesm: call it `ready` and I think it should probably be a buffered driver, with enough space to hold MAX drivers.

pool.go
```diff
@@ -0,0 +1,192 @@
...
+const (
+	DefaultPoolMin               = 1
+	DefaultPoolMax               = 5
+	DefaultPoolTimeBeforeNew     = time.Second * 3
+	DefaultPoolTimeBetweenSpawns = time.Second * 10
+	DefaultPoolTimeBeforeClean   = time.Minute * 1
+)
+
+// DriverPool controls a pool of drivers and balances requests among them,
+// ensuring each driver does not get concurrent requests.
```
alcortesm: Explain that all the drivers in the pool are of the same language.

pool.go
```diff
@@ -0,0 +1,192 @@
...
+type DriverPool struct {
```
alcortesm: I would add an interface for the driver pool, so the main public functionality can be understood.  Then, I would use this as an implementation (with the internal scaler and everything explained in a completely different context than the driver main functionality).  That way people will be able to tell what attributes are there to help fulfilling the interface (example `New`) and which attributes are there to help with the implementation (`TimeBeforeClean`).

pool.go
```diff
@@ -0,0 +1,192 @@
...
+	TimeBeforeNew     time.Duration
+	TimeBetweenSpawns time.Duration
+	TimeBeforeClean   time.Duration
+	New               func() (Driver, error)
+
+	cur     int
+	ch      chan Driver
+	m       *sync.Mutex
+	close   chan struct{}
+	closed  bool
```
alcortesm: this should probably be setted to true in the New method, changed to false in the Start, and checked at the begining of in every other public method.

pool.go
```diff
@@ -0,0 +1,192 @@
...
+	DefaultPoolTimeBetweenSpawns = time.Second * 10
+	DefaultPoolTimeBeforeClean   = time.Minute * 1
+)
+
+// DriverPool controls a pool of drivers and balances requests among them,
+// ensuring each driver does not get concurrent requests.
+type DriverPool struct {
+	// Min
+	Min               int
+	Max               int
```
alcortesm: First thing we have to decide is if Max is going to be constant over the life of a DriverPool.

pool.go
alcortesm: It is important to explain that drivers are reused after they process a request as there are pools that don't reuse their resources.

pool.go
```diff
@@ -0,0 +1,192 @@
...
+func (dp *DriverPool) dec() error {
+	dp.m.Lock()
+	defer dp.m.Unlock()
+	if dp.cur <= dp.Min {
+		logrus.Debugf("cannot decrement instances, already got min: %d", dp.Min)
+		return nil
+	}
+
+	logrus.Debugf("decrement instances")
+	return dp.del(1)
```
alcortesm: Not sure about this, check it out, please:
- del is a blocking operation if there are no ready drivers on the channel.
- At the same time, del is called while holding the lock, this means we will not be able to increment the number of drivers with inc which may break the increment in ParseUast after TimeBeforeNew.

pool.go
```diff
@@ -0,0 +1,192 @@
...
+	if !more {
+		return &protocol.ParseUASTResponse{
+			Status: protocol.Fatal,
+			Errors: []string{"driver pool already closed"},
+		}
+	}
+
+	defer dp.enqueue(d)
+	dp.last = time.Now()
+	return d.ParseUAST(req)
```
abeaumont: May be a good idea to add some kind of timeout to this request, to make sure the driverpool doesn't end up stuck with a bunch of neverending requests?

pool.go
```diff
@@ -0,0 +1,336 @@
...
+// the ScalingPolicy, the timeout for requests before getting a driver assigned,
+// and a function to create driver instances.
+func StartDriverPool(scaling ScalingPolicy, timeout time.Duration, new func() (Driver, error)) (*DriverPool, error) {
+	dp := &DriverPool{
+		scalingPolicy: scaling,
+		timeout:       timeout,
+		new:           new,
+		close:         make(chan struct{}),
+		waiting:       &atomicInt{},
+		readyQueue:    newDriverQueue(1000),
```
alcortesm: 1000 should be a (public?) constant.

pool.go
```diff
@@ -0,0 +1,336 @@
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (dp *DriverPool) scaling() {
+	for {
+		time.Sleep(time.Millisecond * 100)
```
alcortesm: change this to a select on the close chanel and a time.After so this is responsive instead of sleeping.

You will need to change the close channel semantics, but I think that will also be an improvement, in the `Close` method you can send `dp.close <- "please scaling gorutine, die now"` and on this function, the select should react with sending `dp.close <- "scaling is dead for good, you can proceed with the closing"`.

pool.go
```diff
@@ -0,0 +1,336 @@
...
+		s := dp.scalingPolicy.Scale(total, load-ready)
+		_ = dp.setInstanceCount(s)
+	}
+}
+
+// ParseUAST processes a ParseUASTRequest. It gets a driver from the pool and
+// forwards the request to it. If all drivers are busy, it will return an error
+// after the timeout passes. If the DriverPool is closed, an error will be returned.
+func (dp *DriverPool) ParseUAST(req *protocol.ParseUASTRequest) *protocol.ParseUASTResponse {
+	if dp.closed {
```
alcortesm: I'm pretty sure accessing dp.closed here and in `Close` is a data race, maybe https://golang.org/doc/articles/race_detector.html will help to detect this and other similar problems.

pool.go
```diff
@@ -0,0 +1,336 @@
...
+		return nil, true, true
+	}
+}
+
+func (q *driverQueue) Size() int {
+	return int(q.n.Value())
+}
+
+func (q *driverQueue) Close() error {
+
```
alcortesm: remove empty line.

pool.go
```diff
@@ -0,0 +1,336 @@
...
+	atomic.AddInt32(&c.val, int32(n))
+}
+
+func (c *atomicInt) Value() int {
+	return int(atomic.LoadInt32(&c.val))
+}
+
+// ScalingPolicy specifies whether instances should be started or stopped to
+// cope with load.
+type ScalingPolicy interface {
```
alcortesm: This is fucking awesome!

PR #15, opened 2017-05-30
add support for filename and language override

smola:
* Updates to latest protocol version.
* Uses language from the request if provided.
* Uses filename for language detection if provided.


smola:
Requires https://github.com/bblfsh/sdk/pull/118

PR #20, opened 2017-05-31
Support running in Docker

smola:
This PR introduces fixes to be able to run inside a (privileged) Docker container.
Also includes CI setup to push images to Docker Hub.


smola:
So it definitely works: https://hub.docker.com/r/bblfsh/server/tags/ ;)

PR #25, opened 2017-06-14
Set the max number of driver instances to CPU cores

vmarkovtsev:
+ Add the environment variable

### DIFF
--- a/pool.go
+++ b/pool.go
@@ -3,6 +3,9 @@ package server
 import (
 	"fmt"
 	"math"
+	"os"
+	"runtime"
+	"strconv"
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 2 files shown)


pool.go
```diff
@@ -235,7 +239,20 @@ type ScalingPolicy interface {
 // DefaultScalingPolicy returns a new instance of the default scaling policy.
 // Instances returned by this function should not be reused.
 func DefaultScalingPolicy() ScalingPolicy {
-	return MovingAverage(10, MinMax(1, 10, AIMD(1, 0.5)))
+	maxInstances := 0
+	// Try to read maxInstances from the environment variable
+	maxInstancesEnv := os.Getenv("BBLFSH_SCALING_MAX_INSTANCES")
```
abeaumont: I think it'd be better to move this context dependant code outside and pass them as arguments explicitely

vmarkovtsev: I see that `DefaultScalingPolicy()` is used in many places, including the tests. What shall I do with all of those?

smola: @abeaumont that can be done by not using `DefaultScalingPolicy()` and instantiated it instead.

Nonetheless, this code should be in `func init()`, initializing a global variable (e.g. `DefaulMaxInstancesPerDriver`), and then using that variable here.

But you're right that at some point we should move this kind of configuration to `server.go` and propagate it up to the `cli` package. I think would leave that to a further PR though.

vmarkovtsev:
@smola Did as you said, changed the corresponding test to pass.

PR #30, opened 2017-06-15
Prevent potential race condition on drivers map

dennwc:
There is a chance that multiple `ParseUAST` will be invoked simultaneously, thus a data race will occur on `Server.drivers` map.

This change protects map with `RWMutex`.

As a bonus, clients will wait until driver becomes available instead of returning weird errors from RPCs while pull is in progress.

Note: currently server will be blocked with a single driver pull, preventing other language drivers to be pulled concurrently. PR can be extended to add fine-grained locking for each driver pool in the map.

### DIFF
--- a/server.go
+++ b/server.go
@@ -2,6 +2,7 @@ package server
 
 import (
 	"net"
+	"sync"
 
 	"github.com/bblfsh/server/runtime"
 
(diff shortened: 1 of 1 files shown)


smola:
Thanks @dennwc.

Opened this issue for follow up: https://github.com/bblfsh/server/issues/32

PR #35, opened 2017-06-17
Added support for the --encoding parameter en bblfsh client cmd

juanjux:
Small change to support a --encoding parameter in the server. This PR depends on https://github.com/bblfsh/sdk/pull/131.

### DIFF
--- a/cmd/bblfsh/client.go
+++ b/cmd/bblfsh/client.go
@@ -22,6 +22,7 @@ type clientCmd struct {
 	RuntimePath string `long:"runtime-path" description:"runtime path for standalone mode" default:"/tmp/bblfsh-runtime"`
 	ImageRef    string `long:"image" value-name:"image-ref" description:"image reference to use (e.g. docker://bblfsh/python-driver:latest)"`
 	Language    string `long:"language" description:"language of the input" default:""`
+	Encoding    string `long:"encoding" description:"encoding used in the source file" default:"UTF8"`
 	Args        struct {
 		File string `positional-arg-name:"file" required:"true"`
 	} `positional-args:"yes"`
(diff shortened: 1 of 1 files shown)


cmd/bblfsh/client.go
```diff
@@ -39,10 +40,19 @@ func (c *clientCmd) Execute(args []string) error {
 		run = c.runStandalone
 	}
 
+	var encoding protocol.Encoding
+	switch(c.Encoding) {
+	case "Base64":
+		encoding = protocol.Base64
+	default:
+		encoding = protocol.UTF8
```
abeaumont: Since the parameter is optional, I think a better default would be to leave it undefined

juanjux: Well this cover the switch case for "UTF8" and the "no encoding specified" and since the default is UTF8 anyway it's the same result with an ugly empty or missing `default` less.

alcortesm:
> alcortesm approved these changes 3 days ago

:)

cmd/bblfsh/client.go
abeaumont: Sorry, from https://github.com/bblfsh/sdk/pull/131#issuecomment-309488209 I misunderstood that the encoding was optional at the protocol level, but I see in the implementation that it is mandatory, then it doesn't make sense to leave it undefined.

smola:
I'm merging this to keep compatibility with the latest drivers. But let's talk on Monday, I think encoding should be an optional field in the protocol and default to UTF-8.

juanjux:
@smola thats exact how it works.

juanjux:
Closing this one (can't add commits after approval), adding this + the glide.lock fix to github.com/bblfsh/server/pull/45

PR #38, opened 2017-06-26
[fix][vendoring] Rename simple-linguist to enry

juanjux:
Simple rename of simple-linguist to enry in the Glide files and one source file. This actually fix the build of the server for me (previously it gave an error about an "internal reference" to a module), which should not happen since in theory they're the same tags but happens anyway.

### DIFF
--- a/glide.lock
+++ b/glide.lock
@@ -223,7 +223,7 @@ imports:
   - transport
 - name: gopkg.in/src-d/go-errors.v0
   version: 0c303ec4c027302259ec1738f19f515aad25f13f
-- name: gopkg.in/src-d/simple-linguist.v1
+- name: gopkg.in/src-d/enry.v1
   version: 708f2e40bf35cdc6760a11293cad0a8e632b31a1
 - name: gopkg.in/toqueteos/substring.v1
   version: c5f61671513240ddf5563635cc4a90e9f3ae4710
--- a/glide.yaml
+++ b/glide.yaml
@@ -29,7 +29,7 @@ import:
   - http2
 - package: github.com/oklog/ulid
 - package: github.com/Sirupsen/logrus
-- package: gopkg.in/src-d/simple-linguist.v1
+- package: gopkg.in/src-d/enry.v1
 testImport:
 - package: github.com/stretchr/testify
   subpackages:
--- a/language.go
+++ b/language.go
@@ -11,7 +11,7 @@ import (
 func GetLanguage(filename string, content []byte) string {
 	lang := enry.GetLanguage(filename, content)
 	if lang == "" {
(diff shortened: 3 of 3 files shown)


glide.lock
```diff
@@ -223,7 +223,7 @@ imports:
   - transport
 - name: gopkg.in/src-d/go-errors.v0
   version: 0c303ec4c027302259ec1738f19f515aad25f13f
-- name: gopkg.in/src-d/simple-linguist.v1
+- name: gopkg.in/src-d/enry.v1
```
alcortesm: Stupid question: Shouldn't we use `github.com/src-d/enry` instead now that we are versioning using commits?

Now that we are using glide, what we want is to use a particular commit from `github.com/src-d/enry`, not from `gopkg.in/src-d/enry.v1`, right?

ajnavarro: @alcortesm We cannot do this because all the internal enry imports are using gopkg.in/src-d/enry.v1 .

alcortesm: @ajnavarro that's right. But our internal imports use `gopkg.in/src-d/enry.v1` because we were not using glide before and we needed some way to import a particular version.  Now that we are using glide, that should no longer be the case.

What I am talking about is to start importing `github.com/src-d/enry` everywhere, now that thanks to glide, we no longer need to use URL redirects to simulate versioning.

juanjux: To be honest, at this point I don't know what we should or shouldn't be doing with respect to versioning, is there any document or meeting minute detailing this?

smola: We have to use the sale import that the library defines. Enry does use gopkg.in in it's importa, so we do it too. Glide cannot change that.

smola:
@juanjux I've just merged another PR renaming the package, please, rebase.

juanjux:
@smola actually, something is probably odd in enry because if I just import the module without any aliasing and then use enry as the module, the travis build fails with:
```
./language.go:4: imported and not used: "github.com/bblfsh/server/vendor/gopkg.in/src-d/enry.v1" as slinguist
./language.go:11: undefined: enry in enry.GetLanguage
./language.go:13: undefined: enry in enry.OtherLanguage
```
I'll try importing and aliasing as `enry`.

juanjux:
@smola actually did that but without adding the enry alias it searched for the slinguist symbol (wtf) and travis failed, was going to try with the enry alias but I've seen just merged PR #39 just did that so I'm rebasing (the other files in this still need to be modified).

erizocosmico:
@juanjux yep, forgot to update the glide stuff.

glide.lock
alcortesm: :+1:

PR #40, opened 2017-06-26
add JSON REST API

erizocosmico:
/cc @smola 

### Caveats

* Add optional SSL serving?
* Separate command?

### DIFF
--- a/cmd/bblfsh/server.go
+++ b/cmd/bblfsh/server.go
@@ -13,15 +13,11 @@ type serverCmd struct {
 	Address     string `long:"address" description:"server address to bind to" default:"0.0.0.0:9432"`
 	RuntimePath string `long:"runtime-path" description:"runtime path" default:"/tmp/bblfsh-runtime"`
 	Transport   string `long:"transport" description:"default transport to fetch driver images (docker, docker-daemon)" default:"docker"`
+	JSON        bool   `long:"json" description:"start a JSON REST server instead of a gRPC server"`
 }
 
 func (c *serverCmd) Execute(args []string) error {
 	logrus.Debugf("binding to %s", c.Address)
-	//TODO: add support for unix://
-	lis, err := net.Listen("tcp", c.Address)
-	if err != nil {
-		return err
-	}
 
 	r := runtime.NewRuntime(c.RuntimePath)
 	logrus.Debugf("initializing runtime at %s", c.RuntimePath)
@@ -29,8 +25,27 @@ func (c *serverCmd) Execute(args []string) error {
 		return err
 	}
 
-	s := server.NewServer(r)
-	s.Transport = c.Transport
+	if c.JSON {
+		return c.serveJSON(r)
+	}
+
+	return c.serveGRPC(r)
+}
+
+func (c *serverCmd) serveJSON(r *runtime.Runtime) error {
+	s := server.NewRESTServer(r, c.Transport)
+	logrus.Debug("starting server")
+	return s.Serve(c.Address)
+}
+
+func (c *serverCmd) serveGRPC(r *runtime.Runtime) error {
+	//TODO: add support for unix://
+	lis, err := net.Listen("tcp", c.Address)
+	if err != nil {
+		return err
+	}
+
+	s := server.NewGRPCServer(r, c.Transport)
 
 	logrus.Debug("starting server")
 	return s.Serve(lis)
--- a/cmd/bblfsh/server.go
+++ b/cmd/bblfsh/server.go
@@ -13,7 +13,7 @@ type serverCmd struct {
 	Address     string `long:"address" description:"server address to bind to" default:"0.0.0.0:9432"`
 	RuntimePath string `long:"runtime-path" description:"runtime path" default:"/tmp/bblfsh-runtime"`
 	Transport   string `long:"transport" description:"default transport to fetch driver images (docker, docker-daemon)" default:"docker"`
-	JSON        bool   `long:"json" description:"start a JSON REST server instead of a gRPC server"`
+	REST        bool   `long:"rest" description:"start a JSON REST server instead of a gRPC server"`
 }
 
 func (c *serverCmd) Execute(args []string) error {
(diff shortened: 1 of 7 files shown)


abeaumont:
Does it make sense to support two different protocols?

erizocosmico:
@abeaumont we need it for the Babelfish Dashboard. We can't use gRPC on the client (unless we use this alpha-stage lib https://github.com/improbable-eng/grpc-web). Can also be handy for languages without a gRPC library or if you just want JSON representation.

ajnavarro:
Makes more sense for me to have an http server that serves all the data in JSON format that comunicates using gRPC with the bblfsh server.

abeaumont:
@erizocosmico I see the need, but I'm not convinced with having to maintain 2 different protocols, specially since we control both the server and the clients. We can go with @ajnavarro proposal or even evaluate if we really need gRPC at all if a JSON REST API is more universal (I guess the reason for having gRPC is performance, but I'm not convinced the protocol layer is relevant *at all* for performance). We may also end up supporting both protocols aswell, but I think it's worth to discuss the options a bit.

erizocosmico:
The only thing to maintain is exposing the endpoints. It's basically unmarshal request body into X struct, call function, marshal response into json, send response.

I think supporting both protocols is a good thing and makes the service more flexible for the user without much overhead. 

If we take this layer out of here we'd have to put it somewhere else. Keeping it here, forces every change to update the JSON server as well, whereas having it elsewhere would lead to incompatibilities when there are breaking changes here. Not to mention the introduction of another moving part just for the sake of it.

Introducing another layer has a cost: coordination between the version of the two services, maintenance in different places, etc. I just don't see the benefit of having to pay these costs just for not maintaining both, which IMHO has a way lower cost.

abeaumont:
I agree that having another moving part would not make this easier. Not convinced that having 2 protocols is better than having one, even if, as you correctly point, it's not much overhead, but we can try and see how it works, and reevaluate it in the future if needed.

juanjux:
I'm conflicted about this.

I think, and actually said it before, that using JSON(RCP/Rest/Whatever) would make a lot of things easier for the clients (just look at the kludges we've to do in the Python gRPC client, and it's one of the better supported languages used internally in the same company as the authors). 

But... I don't know if we can afford to maintain, test and bugfix two protocols at this early stage of the project. Maybe is something we could discuss in a meeting and then with the community in the channel.

Serabe:
OK, let me clarify this. When we started implementing UAST Viewer we took a look at AST Explorer. We considered implementing just a parser on top of it, but the project has no license and [this issue asking for one](https://github.com/fkling/astexplorer/issues/215) has been opened since April, 3rd 2017 without answer. That means that we needed to reimplement most of it (and by most I mean all) and we did not do it because we were bored, but because what existed could not be used.

We chose protobuffers because is an order of magnitude faster than JSON serialising and deserialising. And by an order of magnitude I really mean it. In my machine, benchmarks showed 3377ns per op, while protobuf is only 160ns, and the struct was way more simpler than a bblfsh UAST.

After doing some benchmark and checking other options, we chose to use protobuf. We invested time (and tears) on making proteus, to make protobuf as simple as possible for Go developers (sadly, at this point in time, they cannot be done simpler). For making easier the implementation of gRPC servers, we invested more time (and tears) on supporting the creation of gRPC services just by adding a `//proteus:generate` before a function. Even methods can be exported to gRPC almost magically (and then people complain about Rails' magic).

@abeaumont (and @juanjux said something similar), I hope you understand that when you say _I'm not convinced with having to maintain 2 different protocols_ one of them is just making sure the proteus' comment is there and proteus is run. Sure it needs a few more lines, both in the Makefile and to create the actual server, but for the server you could just copy them from `proteus/example/server`. For figuring out how protobuf paths work I can do nothing. We invested so much time (and I cried a river) but there is no way to understand them. protobuf team considered documenting them but they rather fix them. In the meantime, neither documentation nor a fix is provided. And I am sorry.

@ajnavarro, there is this little project called https://github.com/grpc-ecosystem/grpc-gateway, and we started adding support for it in proteus (you can see @erizocosmico efforts https://github.com/src-d/proteus/pull/91). The thing is we use gogo/protobuf to improve performance by a lot, and that means not using reflection. It is a well-known limitation of our approach that we cannot do anything that deals with reflection in protobuf.
grpc-gateway uses reflection. We cannot overcome that. Period.

By cannot I don't mean we don't want to, is that we cannot. We can either make your life easier in the general term and let you work with protobuf almost seamlessly while having to implement a small layer for REST API or remove proteus from the equation and let you write your own proto files and compile them. You will need to maintain the protobuf layer, but you would get the REST API for free.

abeaumont:
@Serabe thanks for the insight! As I said above, we can go with both since it's not much overhead (thanks to your work on proteus).
Just a pair of clarifications which while don't change my approval, I think are worth noting:
- Overhead is not only a matter of the code you have to maintain, but of the added complexity in general (like asking which protocol has been used when doing bug triaging, etc.)
- Even if gRPC has an impressive better performance (which I could imagine but didn't know, thanks for the info), I'd bet the (de)serialization time is small in comparison with the parsing and normalization process. This is something that we don't know and should validate though.

cmd/bblfsh/server.go
```diff
@@ -13,24 +13,39 @@ type serverCmd struct {
 	Address     string `long:"address" description:"server address to bind to" default:"0.0.0.0:9432"`
 	RuntimePath string `long:"runtime-path" description:"runtime path" default:"/tmp/bblfsh-runtime"`
 	Transport   string `long:"transport" description:"default transport to fetch driver images (docker, docker-daemon)" default:"docker"`
+	JSON        bool   `long:"json" description:"start a JSON REST server instead of a gRPC server"`
```
smola: Can we make it `--rest`? Looks more clear than just `--json`, which is used in other places in Babelfish for other stuff. Other than that, good to og.

erizocosmico:
Done @smola

Issue #33, opened 2017-06-16
add CLI flag for log level


smola:
Add CLI flag for log level.

abeaumont:
Solved with #41

PR #41, opened 2017-06-27
cmd: Add log-level option

abeaumont:
This solves #33 
Note that the code between client and server is duplicated, but there're already some duplicated options. I think we should add some CommonCmd for these, what do you think?


abeaumont:
I've extracted the shared functionality to its own file. I was wrong and only the log-level parameter is shared so far, since the other parameters are slightly different in client and server code.

PR #50, opened 2017-07-05
Update glide dependencies to latest versions

bzz:
This is mostly motivated to have a version of server that includes latest https://github.com/bblfsh/sdk/pull/135 but it seems to include many more updates.

This is just a result of running `glide up` on the latest master

Shall this be updated to include only bblfsh/sdk to avoid possible consequences?


bzz:
Closing in favor of more localized #50

Issue #29, opened 2017-06-15
Server build fails


dennwc:
Server version: 54c71fc1e313430b3fda16e1e7b0509455cfd5fc
SDK version: https://github.com/bblfsh/sdk/commit/df3e0dadeb24ad49f38561f0955312f3024c82fe

Build log:

```
docker build -f Dockerfile.build -t bblfsh-server-build .
Sending build context to Docker daemon 112.9 MB
Step 1 : FROM golang:1.8-alpine
 ---> e7baf3b1a3a5
Step 2 : RUN apk add --no-cache git make musl-dev musl-utils gcc lvm2-dev btrfs-progs-dev
 ---> Using cache
 ---> fd6daebf7136
Step 3 : ENV GOPATH /go
 ---> Using cache
 ---> 48d0a86e3f57
Step 4 : WORKDIR /go/src/github.com/bblfsh/server
 ---> Using cache
 ---> 4f7d3d5252c2
Successfully built 4f7d3d5252c2
docker run --rm -v /home/dennwc/Go:/go bblfsh-server-build make build-internal
mkdir -p /go/src/github.com/bblfsh/server/build; \
for cmd in bblfsh; do \
        cd /go/src/github.com/bblfsh/server/cmd/${cmd}; \
        go build --ldflags '-X main.version=master -X main.build=06-15-2017_19_51_11' -o /go/src/github.com/bblfsh/server/build/${cmd} .; \
done;
# github.com/bblfsh/server/cmd/bblfsh
/usr/local/go/pkg/tool/linux_amd64/link: running gcc failed: exit status 1
/tmp/go-link-688671004/000001.o: In function `vsnprintf':
/usr/include/x86_64-linux-gnu/bits/stdio2.h:77: undefined reference to `__vsnprintf_chk'
/tmp/go-link-688671004/000001.o: In function `child_func':
/home/dennwc/Go/src/github.com/bblfsh/server/vendor/github.com/opencontainers/runc/libcontainer/nsenter/nsexec.c:212: undefined reference to `__longjmp_chk'
/tmp/go-link-688671004/000001.o: In function `fprintf':
/usr/include/x86_64-linux-gnu/bits/stdio2.h:97: undefined reference to `__fprintf_chk'
/usr/include/x86_64-linux-gnu/bits/stdio2.h:97: undefined reference to `__fprintf_chk'
/usr/include/x86_64-linux-gnu/bits/stdio2.h:97: undefined reference to `__fprintf_chk'
[...]

PR #52, opened 2017-07-05
Fix build failure

bzz:
This is an attempt to address #29 

There are 2 possible fixes so far
 - add `-a` to `go build`
 - ~add `CGO_ENABLED=0`~ `libcontainer` seems to fail the build without it

Let me know if there are other ways to fix it and I will be happy to update PR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -118,7 +118,7 @@ build-internal:
 	mkdir -p $(BUILD_PATH); \
 	for cmd in $(COMMANDS); do \
         cd $(CMD_PATH)/$${cmd}; \
-		$(GO_CMD) build --ldflags '$(LDFLAGS)' -o $(BUILD_PATH)/$${cmd} .; \
+		$(GO_CMD) build -a --ldflags '$(LDFLAGS)' -o $(BUILD_PATH)/$${cmd} .; \
 	done;
 
 docker-image-build: build


abeaumont:
@bzz could you elaborate on how this change fixes the problem?

bzz:
My guess is that it makes the executable linked inside the container.

But this seems to fix symptoms rather then the problem - reasons for #29 are not clear to me.

What do you think?

bzz:
Another approach may be to try to `CGO_ENABLED=0`

abeaumont:
@bzz hmmm, I'd say that having the binary linked inside the container is something expected, isn't it? At the end, that's where it's expected to run. Not sure if `CGO_ENABLED=0` should fix it, if that implies not linking with C library it probably would.
Adding @smola to the discussion in case he's some more knowledge on the topic. Otherwise, I'll merge it since it seems to solve the problem even if we don't understand it well.

bzz:
I have double-checked - disabling CGO at the same line does fix this issue, but build fails in a different way now.

```
        CGO_ENABLED=0 go build --ldflags '-X main.version=master -X main.build=07-05-2017_15_31_18' -o /go/src/github.com/bblfsh/server/build/${cmd} .; \
done;
go build github.com/bblfsh/server/vendor/github.com/opencontainers/runc/libcontainer/nsenter: no buildable Go source files in /go/src/github.com/bblfsh/server/vendor/github.com/opencontainers/runc/libcontainer/nsenter
# github.com/bblfsh/server/vendor/github.com/opencontainers/runc/libcontainer
../../vendor/github.com/opencontainers/runc/libcontainer/init_linux.go:189: undefined: utils.SendFd
make: *** [Makefile:118: build-internal] Error 2
Makefile:115: recipe for target 'build' failed
make: *** [build] Error 2
```

smola:
So I would say that change whatever fix the problem (test produced binary working with and without Docker) and keep an eye on any discussion that might arise here: https://github.com/src-d/ci/pull/34

bzz:
Sounds good. 

@juanjux could you please try running server (built this way) with and _without_ docker?

abeaumont:
It works nice as a standalone binary.

bzz:
Thank you for the prompt reviews guys!

juanjux:
@bzz I tested before inside and outside the docker and in each environment with the ~master, the branch with the updated glide.lock and your fix separately :)

PR #58, opened 2017-07-07
Add a --max-message-size CLI option for both client and server

abeaumont:
Maximum Message size is expressed in MB. Default limit is 4MB. When
limit is defined through this option, it's applied to both directions:
send and receive


abeaumont:
Added some tests

PR #60, opened 2017-07-10
build: Enable image pushing for every commit, tagging as needed

abeaumont:


Makefile
```diff
@@ -61,11 +61,15 @@ ifneq ($(TRAVIS_TAG), )
...
-# if we are not in tag, the push is disabled
-ifeq ($(firstword $(subst -, ,$(DOCKER_VERSION))), $(DOCKER_DEV_PREFIX))
-        pushdisabled = "push disabled for development versions"
+# if we are not in master, the push is disabled
+ifneq ("$(TRAVIS_BRANCH)", "master")
+        pushdisabled = "push disabled for non-master branches"
 endif
 
+# if this is a pull request, the push is disabled
+ifneq ("$(TRAVIS_PULL_REQUEST)", "false")
```
juanjux: `ifeq ("$(TRAVIS_PULL_REQUEST)", "true")` to avoid a double negation? Also I think with the double quotes are not needed in this case (`true`).

abeaumont: There's no `true` value:
* `TRAVIS_PULL_REQUEST`: The pull request number if the current job is a pull request, “false” if it’s not a pull request.

Removed quotes.

PR #61, opened 2017-07-11
changed names related to renaming in sdk/protocol

mcarmonaa:


mcarmonaa:
Travis is failing because of the current sdk/master doesn't contain the renaming changes yet

cmd/bblfsh/client.go
```diff
@@ -53,8 +52,8 @@ func (c *clientCmd) Execute(args []string) error {
 		encoding = protocol.UTF8
 	}
 
-	req := &protocol.ParseUASTRequest{
-		Filename: filepath.Base(c.Args.File),
+	req := &protocol.ParseRequest{
+		Filename: c.Args.File,
```
abeaumont: If I understood it correctly, this line should not change

abeaumont:
@mcarmonaa `glide.lock` should be updated to take the just merged `sdk` version with API changes

glide.lock
```diff
@@ -12,7 +12,7 @@ imports:
   - uast
   - uast/ann
 - name: github.com/BurntSushi/toml
-  version: b26d9c308763d68093482582cea63d69be07a0f0
+  version: a368813c5e648fee92e5f6c30e3944ff9d5e8895
```
juanjux: Mmm, I'm not sure we should update all dependencies to their `~master` every time we update the SDK one. We probably should just update to the latest releases, trough that could depend on every dependency.

abeaumont: Agree. This should be fixed with proper versioning.

abeaumont:
@mcarmonaa travis is failing due to `enry`, please take a look at it.

glide.lock
mcarmonaa: note that set a version to master downloads the latest release, not the real last commit in master :S

PR #62, opened 2017-07-11
build: Fix shell syntax

abeaumont:


Makefile
```diff
@@ -142,6 +142,6 @@ push: docker-image-build
 	$(DOCKER_PUSH) $(call unescape_docker_tag,$(DOCKER_IMAGE_VERSIONED))
 	@if [ "$$TRAVIS_TAG" != "" ]; then \
 		$(DOCKER_TAG) $(call unescape_docker_tag,$(DOCKER_IMAGE_VERSIONED)) \
-			$(call unescape_docker_tag,$(DOCKER_IMAGE)):latest
-		$(DOCKER_PUSH) $(call unescape_docker_tag,$(DOCKER_IMAGE):latest)
+			$(call unescape_docker_tag,$(DOCKER_IMAGE)):latest; \
```
juanjux: Shouldn't this `;` be an `&&`?

abeaumont: It can be both, I used `;` because:
* it's the way it was, and has worked fine so far
* it'll fail eitherwise
* it may give a false sensation of having some kind of transactional support, which we don't, and I don't think it's worth at this moment.

I'm not strongly against it though, so I can change it if you prefer.

juanjux: I'm not string against it either, just commenting the usual way to write those commands in Dockerfiles. Your arguments make sense and the only advantage of the `&&` would be that the script would fail at the specific point of failure (and not one command later) so leave it with the `;` if you want.

PR #65, opened 2017-07-12
Add profiler support

bzz:
### Client

Profiler support to client, save results to FS

 Test plan:
 - start server as usual `sudo bblfsh server`
 - start client with new flag
   `bblfsh client --cpuprofile ./sample.pprof sample.py`
 - `go tool pprof ./sample.pprof`

 ### Server

Profiler support to server, expose results over HTTP

Test plan:
```
sudo server --profiler
go tool pprof http://localhost:6062/debug/pprof/profile
```

```
sudo server --profiler --rest
go tool pprof http://localhost:6062/debug/pprof/profile
```

### DIFF
--- a/cmd/bblfsh/client.go
+++ b/cmd/bblfsh/client.go
@@ -7,6 +7,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"runtime/pprof"
 
 	"github.com/bblfsh/server"
 	"github.com/bblfsh/server/runtime"
@@ -24,12 +25,45 @@ type clientCmd struct {
 	ImageRef    string `long:"image" value-name:"image-ref" description:"image reference to use (e.g. docker://bblfsh/python-driver:latest)"`
 	Language    string `long:"language" description:"language of the input" default:""`
 	Encoding    string `long:"encoding" description:"encoding used in the source file" default:"UTF8"`
+	CpuProfile  string `short:"" long:"cpuprofile" description:"path to file where Cpu Profile will be stored" default:""`
+	MemProfile  string `short:"" long:"memprofile" description:"path to file where Memory Profile will be stored" default:""`
 	Args        struct {
 		File string `positional-arg-name:"file" required:"true"`
 	} `positional-args:"yes"`
 }
 
+func (c *clientCmd) StartCpuProfilingMaybe() {
+	if c.CpuProfile != "" {
+		f, err := os.Create(c.CpuProfile)
+		if err != nil {
+			logrus.Errorf("Failed to create a CpuProfile file at %s, err:%s", c.CpuProfile, err)
+			os.Exit(1)
+		}
+		logrus.Infof("Start CPU profiling, save to %s", c.CpuProfile)
+		pprof.StartCPUProfile(f)
+	}
+}
+func (c *clientCmd) StopCpuProfilingMaybe() {
+	logrus.Info("Stop CPU profiling")
+	pprof.StopCPUProfile()
+}
+
+// If profiling enabled though CLI, it saves memory profile to file.
+func (c *clientCmd) SaveMemProfileMaybe() {
+	if c.MemProfile != "" {
+		f, err := os.Create(c.MemProfile)
+		if err != nil {
+			logrus.Errorf("Failed to save Heap profile to %s, err:%s", c.MemProfile, err)
+			os.Exit(1)
+		}
+		logrus.Infof("Save Heap profile to %s", c.MemProfile)
+		pprof.WriteHeapProfile(f)
+		f.Close()
+	}
+}
+
 func (c *clientCmd) Execute(args []string) error {
+	c.StartCpuProfilingMaybe()
 	if err := c.exec(args); err != nil {
 		return err
 	}
@@ -64,7 +98,11 @@ func (c *clientCmd) Execute(args []string) error {
 		return err
 	}
 
+	c.StopCpuProfilingMaybe()
+	c.SaveMemProfileMaybe()
 	prettyPrinter(os.Stdout, resp)
+
+	c.StopCpuProfilingMaybe()
 	return nil
 }
 
--- a/cmd/bblfsh/client.go
+++ b/cmd/bblfsh/client.go
@@ -25,45 +25,54 @@ type clientCmd struct {
 	ImageRef    string `long:"image" value-name:"image-ref" description:"image reference to use (e.g. docker://bblfsh/python-driver:latest)"`
 	Language    string `long:"language" description:"language of the input" default:""`
 	Encoding    string `long:"encoding" description:"encoding used in the source file" default:"UTF8"`
-	CpuProfile  string `short:"" long:"cpuprofile" description:"path to file where Cpu Profile will be stored" default:""`
-	MemProfile  string `short:"" long:"memprofile" description:"path to file where Memory Profile will be stored" default:""`
+	CPUProfile  string `long:"cpuprofile" description:"path to file where Cpu Profile will be stored" default:""`
+	MemProfile  string `long:"memprofile" description:"path to file where Memory Profile will be stored" default:""`
 	Args        struct {
 		File string `positional-arg-name:"file" required:"true"`
 	} `positional-args:"yes"`
+	CPUProfileFile *os.File
 }
 
-func (c *clientCmd) StartCpuProfilingMaybe() {
-	if c.CpuProfile != "" {
-		f, err := os.Create(c.CpuProfile)
+func (c *clientCmd) StartCPUProfileMaybe() error {
+	if c.CPUProfile != "" {
+		f, err := os.Create(c.CPUProfile)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


cmd/bblfsh/client.go
```diff
@@ -24,12 +25,45 @@ type clientCmd struct {
 	ImageRef    string `long:"image" value-name:"image-ref" description:"image reference to use (e.g. docker://bblfsh/python-driver:latest)"`
 	Language    string `long:"language" description:"language of the input" default:""`
 	Encoding    string `long:"encoding" description:"encoding used in the source file" default:"UTF8"`
+	CpuProfile  string `short:"" long:"cpuprofile" description:"path to file where Cpu Profile will be stored" default:""`
```
abeaumont: We aren't specifying empty short versions. I'd do the same for consistency.

cmd/bblfsh/client.go
```diff
@@ -24,12 +25,45 @@ type clientCmd struct {
...
+	pprof.StopCPUProfile()
+}
+
+// If profiling enabled though CLI, it saves memory profile to file.
+func (c *clientCmd) SaveMemProfileMaybe() {
+	if c.MemProfile != "" {
+		f, err := os.Create(c.MemProfile)
+		if err != nil {
+			logrus.Errorf("Failed to save Heap profile to %s, err:%s", c.MemProfile, err)
+			os.Exit(1)
```
abeaumont: Return the error instead of exiting.

cmd/bblfsh/client.go
```diff
@@ -24,12 +25,45 @@ type clientCmd struct {
...
+		}
+		logrus.Infof("Save Heap profile to %s", c.MemProfile)
+		pprof.WriteHeapProfile(f)
+		f.Close()
```
abeaumont: may be better to `defer` it?

cmd/bblfsh/client.go
```diff
@@ -24,12 +25,45 @@ type clientCmd struct {
...
 		File string `positional-arg-name:"file" required:"true"`
 	} `positional-args:"yes"`
 }
 
+func (c *clientCmd) StartCpuProfilingMaybe() {
+	if c.CpuProfile != "" {
+		f, err := os.Create(c.CpuProfile)
+		if err != nil {
+			logrus.Errorf("Failed to create a CpuProfile file at %s, err:%s", c.CpuProfile, err)
+			os.Exit(1)
```
abeaumont: Return the error instead of exiting.

cmd/bblfsh/client.go
```diff
@@ -24,12 +25,45 @@ type clientCmd struct {
...
+		}
+		logrus.Infof("Start CPU profiling, save to %s", c.CpuProfile)
+		pprof.StartCPUProfile(f)
```
abeaumont: Forgot to close `f`. Maybe defer it?

cmd/bblfsh/client.go
```diff
@@ -24,12 +25,45 @@ type clientCmd struct {
...
+	}
+}
+func (c *clientCmd) StopCpuProfilingMaybe() {
```
abeaumont: * Not sure the `Maybe` is appropiate here, since you're calling the `StopCPUProfile` unconditionally.
* `StopCPU` would be more consistent with called function name.

cmd/bblfsh/client.go
```diff
@@ -64,7 +98,11 @@ func (c *clientCmd) Execute(args []string) error {
 		return err
 	}
 
+	c.StopCpuProfilingMaybe()
+	c.SaveMemProfileMaybe()
 	prettyPrinter(os.Stdout, resp)
+
+	c.StopCpuProfilingMaybe()
```
abeaumont: Does it make sense to stop the profiler twice?

cmd/bblfsh/server.go
```diff
@@ -19,10 +20,12 @@ var (
...
 	commonCmd
-	Address     string `long:"address" description:"server address to bind to" default:"0.0.0.0:9432"`
-	RuntimePath string `long:"runtime-path" description:"runtime path" default:"/tmp/bblfsh-runtime"`
-	Transport   string `long:"transport" description:"default transport to fetch driver images (docker, docker-daemon)" default:"docker"`
-	REST        bool   `long:"rest" description:"start a JSON REST server instead of a gRPC server"`
+	Address      string `long:"address" description:"server address to bind to" default:"0.0.0.0:9432"`
+	RuntimePath  string `long:"runtime-path" description:"runtime path" default:"/tmp/bblfsh-runtime"`
+	Transport    string `long:"transport" description:"default transport to fetch driver images (docker, docker-daemon)" default:"docker"`
+	REST         bool   `long:"rest" description:"start a JSON REST server instead of a gRPC server"`
+	Profiler     bool   `long:"profiler" description:"start CPU & memeory profiler"`
```
abeaumont: typo in `memeory`

cmd/bblfsh/server.go
```diff
@@ -2,15 +2,16 @@ package main
 
 import (
 	"net"
+	"net/http"
+	_ "net/http/pprof"
```
abeaumont: Excuse my ignorance, what's this underscore for?

bzz: I by no means am an expert, but I believe this to be an idiom called `blank identifier` https://golang.org/doc/effective_go.html#blank

In `import` context it prevents a compiler from complaining on un-used imports. I believe here such import have a side-effect of enabling profiling instrumentation.

bzz:
>  but I'm eager to see what we can learn from this profiling!

@abeaumont there is a teaser (full graph in .svg and top10 hot functions) of server profiling results at https://gist.github.com/bzz/f7adb8abd32502998c3427179742c468#performance-of-data-collection-pipeline

bzz:
Reviews addressed in 0a1dd65

cmd/bblfsh/server.go
abeaumont: Understood, thanks!

cmd/bblfsh/client.go
```diff
@@ -25,45 +25,54 @@ type clientCmd struct {
...
-func (c *clientCmd) SaveMemProfileMaybe() {
+func (c *clientCmd) SaveMemProfileMaybe() error {
 	if c.MemProfile != "" {
 		f, err := os.Create(c.MemProfile)
 		if err != nil {
-			logrus.Errorf("Failed to save Heap profile to %s, err:%s", c.MemProfile, err)
-			os.Exit(1)
+			errFmt := "Failed to save Heap profile to %s, err:%s"
+			logrus.Errorf(errFmt, c.MemProfile, err)
+			return fmt.Errorf(errFmt, c.MemProfile, err)
```
abeaumont: Use `go-errors` for consistency, I think `Wrap` may be handy here

abeaumont:
@bzz nice teaser! Thanks for sharing!

bzz:
@abeaumont `go-errors` added in d662298

cmd/bblfsh/client.go
```diff
@@ -15,6 +15,11 @@ import (
 	"github.com/Sirupsen/logrus"
 	"github.com/bblfsh/sdk/protocol"
 	"google.golang.org/grpc"
+	"srcd.works/go-errors.v0"
+)
+
+var (
+	errProfiler = errors.NewKind("Failed to create a % file at %s")
```
abeaumont: Capitialize `errProfiler` -> `ErrProfiler`

bzz: I did not export this var intentionally.
Do you see any cases for this error to be useful outside of the current package?

abeaumont: Good point, but this is causing the build error, so change the name of the variable where used then 🙂

abeaumont: I've been reading a bit more about error handling in Golang, and I'm not so sure that variable should be private. I went through the code to look at how we're using errors and to mark some more as private, and also read https://github.com/golang/go/wiki/Errors#naming, and I think that if someone was to use this package externally (which it doesn't make much sense itself, but assuming it would), since this error is returned from a public function, it may make sense for the user to try to check if the error is a `ErrProfiler` for some reason, doesn't it?

bzz: Well, yes, I you are right, if someone was to import it - it would make sense to expose possible recoverable error conditions. And because it's not really an error condition at all, previous impl did not pass it out of the function.

I can hardly imagine anyone willing to import `main` package as dependency (IDK if that is possible), but as soon as we propagate this error, I guess you are right and we should expose it.

bzz:
History re-written, feedback addressed in 353e063

cmd/bblfsh/client.go
abeaumont: I agree that exposing anything in `main` doesn't seem to make sense, and I'll take a look at why is that and change it if needed, but that goes beyond this PR, so I'd expose for consistency for now.

abeaumont:
Thanks!

PR #67, opened 2017-07-13
set version for dependencies

mcarmonaa:
* versions are set to allow minor changes following semantic version, (in the end the specific versions are in glide.lock)
* since bblfsh/sdk and containers/images don't have releases yet, their versions have been set to the current last commit

It passes the tests with this versions.

### DIFF
--- a/glide.yaml
+++ b/glide.yaml
@@ -1,39 +1,56 @@
 package: github.com/bblfsh/server
 import:
+- package: github.com/Sirupsen/logrus
+  version: ^1.0.2
 - package: github.com/bblfsh/sdk
+  version: 9db6b2ee9500857b54f8d40165dc92aba482aadf
   subpackages:
   - manifest
   - protocol
+  - protocol/driver
 - package: github.com/containers/image
   version: 66efd5c31ce9470c37d8fb40c2e424c63f72c735
   subpackages:
+  - directory
   - docker
+  - docker/archive
+  - docker/daemon
   - image
-  - manifest
-  - types
+  - oci/layout
+  - ostree
   - transports
+  - types
+- package: github.com/gin-gonic/gin
+  version: ^1.2.0
+- package: github.com/jessevdk/go-flags
+  version: ^1.2.0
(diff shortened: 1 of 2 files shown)


mcarmonaa:
Since I can't assign reviewers, I mention the maintainer: @abeaumont

glide.yaml
```diff
@@ -1,39 +1,56 @@
 package: github.com/bblfsh/server
 import:
+- package: github.com/Sirupsen/logrus
+  version: ^1.0.2
 - package: github.com/bblfsh/sdk
+  version: 9db6b2ee9500857b54f8d40165dc92aba482aadf
```
abeaumont: can't this be master?

glide.yaml
```diff
@@ -1,39 +1,56 @@
...
-  version: v1.3.0
+  version: ^0.8.0
 - package: github.com/rs/cors
   version: ^1.1.0
+- package: google.golang.org/grpc
+  version: ^1.4.2
+- package: gopkg.in/src-d/enry.v1
+  version: ^1.3.0
+- package: gopkg.in/src-d/go-errors.v0
+  version: 6f07baca276330431b44ece8ee84ac98167bc4ea
```
abeaumont: same here

glide.yaml
```diff
@@ -1,39 +1,56 @@
...
+- package: srcd.works/go-errors.v0
+  version: 6f07baca276330431b44ece8ee84ac98167bc4ea
```
abeaumont: same here

glide.yaml
mcarmonaa: it could be, but master depends on the cache, it could use a commit previous to this one if `glide` cached it when master was at this point.

In this way you are sure to use this specific commit and if you have to run `glide up` you wouldn't have problems with this cache/version issues

abeaumont: but a `glide up` wouldn't update this file, so we have to track specific commits by hand, is that correct?

abeaumont: as commented offiline, that's the problem.

PR #71, opened 2017-07-14
Use server as entrypoint to allow pass options

mcarmonaa:


mcarmonaa:
Since I can't assign reviewers yet, I mention to mantainer: @abeaumont

Issue #75, opened 2017-07-19
Reduce default verbosity of CLI logs


bzz:
Processing 1000s or repos results in 100+mb of server logs mostly due to UASTs printed to stdout.
It can be avoided by skipping printing UASTs to stdout.

May be it could be hidden in `-v` or `-vv`   or some deeper log-level? As it is very useful for lower-level debugging.

abeaumont:
You're right @bzz, thanks for reporting. I'll take a look and get back with a proposal.

abeaumont:
@bzz for some reason I cannot add you to #76 as a reviewer, but please take a look. I've just categorized better current log messages, which should be enough I think for large scale needs. I'm open to suggestions for better fine-tuning this messages, so please let me know if you miss anything relevant at the `info` level for normal execution at scale.

bzz:
Sorry, I must have missed it. 
I think #76 addresses this issue perfectly

abeaumont:
Great, closing then. Should you find any way to improve default logging please open a new issue or PR

PR #76, opened 2017-07-20
Improve logging

abeaumont:
Try to categorize log messages better:
* Move level from `info` to `debug` in some messages, leaving `debug` really for debugging information
* Set `info` as default log level.

Attempts to solve #75


bzz:
Looks awesome to me, thank you @abeaumont !

PR #85, opened 2017-08-10
build: Make bblfsh binary static

### LINKED ISSUE #83: released bblfsh binary is dynamically linked against musl
```
$ wget https://github.com/bblfsh/server/releases/download/v0.6.0/bblfsh
$ ldd bblfsh 
	linux-vdso.so.1 =>  (0x00007ffc52dac000)
	libc.musl-x86_64.so.1 => not found
```

cc @rporres

abeaumont:
Fixes #83

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,7 +33,7 @@ ifneq ($(origin TRAVIS_TAG), undefined)
 endif
 
 # Build
-LDFLAGS = -X main.version=$(BRANCH) -X main.build=$(BUILD)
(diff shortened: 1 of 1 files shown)


juanjux:
And here I was thinking that Go compiles statically by default...

While reading about this I also [found that somebody suggest](https://stackoverflow.com/questions/33113190/how-to-create-a-statically-linked-golang-executable-with-go-1-5) to also add `-linkmode external` for Go > 1.5 (untested).

abeaumont:
We are already using Go > 1.5 and seems to work fine, but I've found lots of confusing information about this topic, so we may need to improve this in the future. Merging this for now.

PR #95, opened 2017-08-29
Return empty Response for empty Request, avoiding instantiating drivers

juanjux:
Part of src-d/backlog#889

### DIFF
--- a/server.go
+++ b/server.go
@@ -122,7 +122,7 @@ func (s *Server) Parse(req *protocol.ParseRequest) *protocol.ParseResponse {
 
 	// If the code Content is empty, just return an empty reponse
 	if req.Content == "" {
-		logrus.Info("Empty code received, returning empty UAST")
+		logrus.Debug("Empty code received, returning empty UAST")
 		return &protocol.ParseResponse{
 			Status: protocol.Ok,
 			UAST: &uast.Node{},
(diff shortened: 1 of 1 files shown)


juanjux:
Yes, definitely, I'll update the docs. Another option would be to add a new response status `empty` or `warning` (the latter using `[]errors` to include the warning).

juanjux:
Don't merge yet - I wan't to explore options to return an empty `Response.UAST` field instead of `nil`.

bzz:
Just 5c - it would if that would have no implications on error handling for the clients

juanjux:
I changed the response to have an empty `Node` in its `ParseResponse.UAST` field instead of `nil`. This should make it less prone to cause unwanted `nil` dereferences in user code (even trough I think users should check any pointers against `nil` before dereferencing, in general) and have a better correspondence (empty code -> empty UAST).

PTAL and merge if approved.

server.go
```diff
@@ -119,6 +120,15 @@ func (s *Server) Parse(req *protocol.ParseRequest) *protocol.ParseResponse {
 		lang = GetLanguage(req.Filename, []byte(req.Content))
 	}
 
+	// If the code Content is empty, just return an empty reponse
+	if req.Content == "" {
+		logrus.Info("Empty code received, returning empty UAST")
```
abeaumont: Not very relevant information about the server operation itself, maybe a `Debug` is better?

PR #100, opened 2017-09-12
dependencies: Remove usage of srcd.works

### LINKED ISSUE #99: Remove usage of obsolete package "srcd.works/go-errors.v0"
CI has failed due to srcd.works not existing anymore: https://travis-ci.org/bblfsh/server/builds/273847022
It should be replaced by `gopkg.in/src-d/go-errors.v0`

abeaumont:
Fixes #99

### DIFF
--- a/cmd/bblfsh/client.go
+++ b/cmd/bblfsh/client.go
@@ -15,7 +15,7 @@ import (
 	"github.com/Sirupsen/logrus"
 	"google.golang.org/grpc"
 	"gopkg.in/bblfsh/sdk.v0/protocol"
-	"gopkg.in/src-d/go-errors.v0"
+	"gopkg.in/src-d/go-errors.v1"
 )
 
 var (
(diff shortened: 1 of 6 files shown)


cmd/bblfsh/client.go
```diff
@@ -15,7 +15,7 @@ import (
 	"github.com/Sirupsen/logrus"
 	"google.golang.org/grpc"
 	"gopkg.in/bblfsh/sdk.v0/protocol"
-	"srcd.works/go-errors.v0"
+	"gopkg.in/src-d/go-errors.v0"
```
juanjux: Just wanted to comment that there is a `v1` too, linking to `v1.0.0`, I'm not sure if we must use `v0`.

abeaumont: The reason for using `v0` is that we are already using it in other parts of the repo, but looking at the difference betweent the two, there are none, so I'll upgrade to `v1`.

abeaumont:
Upgraded to v1

PR #102, opened 2017-09-14
[Do not merge] Upgrade to SDK v1

abeaumont:
Upgrades server to use SDK v1, but there's not point of merging this till we have drivers working on v1 aswell (at this point I don't think it's worth the extra work of maintaining different branches).


juanjux:
I guess we can merge this one to test the drivers and clients against this version.

abeaumont:
Let me get the go client ready first and I'll merge it.

PR #104, opened 2017-09-28
*: implementation of grpc drivers

mcuadros:
This is the implementation of the grpc driver, it's working, but it a preliminary version. I should work in logging (server and from driver), cleanup of the code and error handling (mainly for the old drivers)

PTAL to the architecture, the code may contain leftovers and/or quick and dirty stuff.


driver.go
```diff
@@ -1,81 +1,142 @@
...
+	Start() error
+	Stop() error
+	Service() protocol.ProtocolServiceClient
+}
+
+type DriverInstance struct {
+	ctx  context.Context
+	p    *runtime.Process
+	c    runtime.Container
+	i    runtime.DriverImage
```
abeaumont: `proc` and `img` may be better names than `p` and `i`.

driver.go
```diff
@@ -1,81 +1,142 @@
...
+	if err := i.c.Start(); err != nil {
+		logrus.Errorf("error starting container %s (%s): %s", i.i.Name(), i.c.ID(), err)
+		return err
+	}
+
+	if err := i.dial(); err != nil {
+		return err
+	}
+
+	if err := i.loadVersion(); err != nil {
```
abeaumont: What is this for?

mcuadros: is a validation of the connection.

driver.go
mcuadros: ok

mcuadros:
done @juanjux @abeaumont

Makefile
```diff
@@ -88,7 +90,8 @@ $(NOVENDOR_PACKAGES):
 	$(GO_GET) $@
 
 $(VENDOR_PATH):
-	$(GLIDE) install
+	$(GLIDE) install; \
+	rm -rf vendor/github.com/Sirupsen/;
```
abeaumont: Why is this needed?

cmd/bblfshd/main.go
```diff
@@ -0,0 +1,158 @@
...
+	overrides := make(map[string]string)
+	for _, img := range strings.Split(os.Getenv("BBLFSH_DRIVER_IMAGES"), ";") {
+		if img = strings.TrimSpace(img); img == "" {
+			continue
+		}
+
+		fields := strings.Split(img, "=")
+		if len(fields) != 2 {
+			logger.Errorf("invalid image driver format %s", img)
+			os.Exit(1)
```
abeaumont: Having so many exit points makes code hard to reuse/test. It may be better to just return the error and let the caller handle it properly.

glide.lock
```diff
@@ -166,10 +174,10 @@ imports:
   version: 7af7a1e09ba336d2ea14b1ce73bf693c6837dbf6
 - name: github.com/seccomp/libseccomp-golang
   version: f6ec81daf48e41bf48b475afc7fe06a26bfb72d1
-- name: github.com/Sirupsen/logrus
-  version: f006c2ac4710855cf0f916dd6b77acf6b048dc6e
 - name: github.com/sirupsen/logrus
   version: 89742aefa4b206dcf400792f3bd35b542998eb3b
+- name: github.com/Sirupsen/logrus
```
abeaumont: Do we need to have these duplicated?

glide.yaml
```diff
@@ -37,7 +37,8 @@ import:
 - package: google.golang.org/grpc
   version: ^1.4.2
 - package: gopkg.in/bblfsh/sdk.v1
-  version: ^1.0.0-rc1
+  version: 83b3b64de663069fd7e0a12d08db4c00da3c17ce
```
abeaumont: This should be reverted back to a named version once SDK changes are released.

cmd/bblfshd/main.go
mcuadros: Witch caller, is a main

Makefile
mcuadros: https://github.com/sirupsen/logrus/issues/570#issuecomment-313933276
Will be deleted soon.

glide.lock
mcuadros: Yep, sadly https://github.com/sirupsen/logrus/issues/570#issuecomment-313933276 and some projects are using the other one. My next PR will be updating all the dependencies.

cmd/bblfshd/main.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+
+	logger server.Logger
+)
+
+func init() {
+	cmd := flag.NewFlagSet("server", flag.ExitOnError)
+	network = cmd.String("network", "tcp", "network type: tcp, tcp4, tcp6, unix or unixpacket.")
+	address = cmd.String("address", "localhost:9432", "address to listen.")
+	storage = cmd.String("storage", "/var/lib/bblfshd", "path where all the runtime information is stored.")
```
juanjux: It would be nice to have an option to not cache the images; it would be very useful when you're debugging interactions between drivers and the servers and you change the local image frequently. But that's outside the scope of this PR and could be done later.

cmd/bblfshd/main.go
abeaumont: `main` function is the caller. I guess this is a matter of taste, since it's a direct call from main.

mcuadros: I just wanted to make the main function simple and avoid error handling there...

cmd/bblfshd/main.go
mcuadros: The images should be stored and updated if need it. A feature to be develop is a command line to to force this update in the server, as `docker pull <image>`

runtime/container_test.go
```diff
@@ -231,7 +232,7 @@ func TestContainerStartWaitExit1(t *testing.T) {
 		Stdout: out,
 	}
 
-	c, err := rt.Container(d, p)
+	c, err := rt.Container("start-wait-exit1", d, p, nil)
```
juanjux: exit1?

runtime/runtime.go
```diff
@@ -66,9 +67,15 @@ func (r *Runtime) ListDrivers() ([]*DriverImageStatus, error) {
 }
 
 // Container returns a container for the given DriverImage and Process
-func (r *Runtime) Container(d DriverImage, p *Process) (Container, error) {
+func (r *Runtime) Container(id string, d DriverImage, p *Process, f ConfigFactory) (Container, error) {
+	//id := NewULID().String()
```
juanjux: Remove?

daemon/daemon.go
```diff
@@ -0,0 +1,260 @@
...
+	transport := s.Transport
+	if transport == "" {
+		transport = defaultTransport
+	}
+
+	ref := fmt.Sprintf("bblfsh/%s-driver:latest", lang)
+	switch transport {
+	case "docker":
+		ref = "//" + ref
+	}
```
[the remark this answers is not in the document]
mcuadros: this is old code, and planned to be deprecated.

daemon/daemon.go
```diff
@@ -0,0 +1,260 @@
...
+
+func getDriverInstanceOptions(logger server.Logger) *Options {
+	opts := &Options{}
+
+	var l *logrus.Logger
+	switch i := logger.(type) {
+	case *logrus.Logger:
+		l = i
+	case *logrus.Entry:
+		l = i.Logger
```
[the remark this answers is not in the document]
mcuadros: I was thinking about this, I could find nothing better, without mixing concepts.

runtime/container_test.go
mcuadros: its the name of the test :/

runtime/runtime.go
mcuadros: sure

runtime/container_test.go
juanjux: Ahhh, didn't notice that.

cmd/bblfshd/main.go
```diff
@@ -0,0 +1,158 @@
...
+		overrides[lang] = image
+	}
+
+	return overrides
+}
+
+func handleGracefullyShutdown(l net.Listener, d *daemon.Daemon) {
+	var gracefulStop = make(chan os.Signal)
+	signal.Notify(gracefulStop, syscall.SIGTERM)
+	signal.Notify(gracefulStop, syscall.SIGINT)
```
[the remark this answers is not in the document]
mcuadros: You have many signal to catch like SIGTERM, SIGINT, SIGQUIT, SIGABRT, SIGKILL, but what I said are more or less the more commons, I prefer keep does, and see if they some others a missing.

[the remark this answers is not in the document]
juanjux: Only Jesus can catch SIGKILL :) other than that I'm Ok with handling these two.

PR #105, opened 2017-09-29
*: go packages, go version and dockerfile updated

mcuadros:
Since runc and other dependencies are pretty new and go 1.9 was released, is interesting to use the latest version to avoid old bugs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,7 @@
 language: go
 
 go:
-  - 1.8
-  - tip
-
-matrix:
-  fast_finish: true
-  allow_failures:
-    - go: tip
+  - 1.9
 
 sudo: required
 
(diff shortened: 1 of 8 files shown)


.travis.yml
```diff
@@ -1,20 +1,25 @@
...
 
 go:
-  - 1.8
-  - tip
-
-matrix:
-  fast_finish: true
-  allow_failures:
-    - go: tip
+  - 1.9
```
juanjux: Maybe we should continue testing on Go 1.8 (and 1.9), it's still the default version for Ubuntu and other distros.

mcuadros: This is a compiled program, we compile for 1.9 and the binaries and docker are release to with version.

.travis.yml
```diff
@@ -1,20 +1,25 @@
...
 
 services:
   - docker
 
 before_install:
+  - sudo apt-get -qq update
+  - sudo apt-get -qq install btrfs-tools libdevmapper-dev libgpgme11-dev libapparmor-dev libseccomp-dev
+  - sudo apt-get -qq install autoconf automake bison e2fslibs-dev libfuse-dev libtool liblzma-dev gettext
+  - OSTREE_VERSION=v2017.9
+  - git clone https://github.com/ostreedev/ostree ${TRAVIS_BUILD_DIR}/ostree
```
juanjux: This dependency should probably be documented in the README build section.

.travis.yml
juanjux: It's still useful to automate testing with travis for common go versions in case somebody wants to build the server.

PR #112, opened 2017-10-04
runtime: ImageConfig, including original image reference and changed storage schema

mcuadros:
The json saved together the image, now includes the original image reference.
Also the paths of the images are based on a digest and not in the raw name of the image to avoid problem with slashes.


abeaumont:
Didn't see that the change was actually _removing_ them :)

PR #113, opened 2017-10-05
bblfshctl: new commands parse and drivers

mcuadros:
It's implements a two new commands:
- parse: to parse files and prints the UAST or the AST, mainly for testing and demoing
- drivers: list the installed drivers.

### DIFF
--- a/cli/bblfshctl/cmd/base.go
+++ b/cli/bblfshctl/cmd/base.go
@@ -7,27 +7,46 @@ import (
 	"github.com/bblfsh/server/daemon/protocol"
 
 	"google.golang.org/grpc"
+	sdk "gopkg.in/bblfsh/sdk.v1/protocol"
 )
 
-type GRPCCommand struct {
-	Network string `long:"endpoint" default:"unix" description:"control server network type"`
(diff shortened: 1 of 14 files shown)


cli/bblfshctl/cmd/base.go
```diff
@@ -7,27 +7,46 @@ import (
...
+type UserCommand struct {
+	Network string `long:"endpoint" default:"tcp" description:"server network type"`
+	Address string `long:"address" default:"localhost:9432" description:"server address to connect"`
+
+	conn *grpc.ClientConn
+	srv  sdk.ProtocolServiceClient
+}
+
+func (c *UserCommand) Execute(args []string) error {
+	conn, err := dialGRPC(c.Network, c.Address)
```
abeaumont: Slightly simpler:

```go
c.conn, err = ...
c.srv = sdk....(c.conn)
```

mcuadros: the same number of line :p

abeaumont: It should be a line less, but yes, it's a _slight_ improvement =)

PR #137, opened 2017-11-30
[README] Add macOS command

juanjux:
Signed-off-by: Juanjo Alvarez <juanjo@sourced.tech>

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,13 @@ The recommended way to run *bblfshd* is using *docker*:
 docker run -d --name bblfshd --privileged -p 9432:9432 -v /var/lib/bblfshd:/var/lib/bblfshd bblfsh/bblfshd
 ```
 
+On macOS, use this command instead to use a docker volume:
+
+```sh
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -20,6 +20,13 @@ The recommended way to run *bblfshd* is using *docker*:
 docker run -d --name bblfshd --privileged -p 9432:9432 -v /var/lib/bblfshd:/var/lib/bblfshd bblfsh/bblfshd
 ```
 
+On macOS, use this command instead to use a docker volume:
+
+```sh
+docker run -d --name bblfshd --privileged -p 9432:9432 -v bblfsh-cache:/var/lib/bblfshd bblfsh/bblfshd
```
mcuadros: cache? why is called like this?

juanjux: Well is a cache of driver images by I don't care much about the name, what do you propose?

mcuadros: Is not a cache, is the storage, of the containers and drivers.

juanjux: Oooook, PR updated.

PR #150, opened 2018-02-22
Migrate from Glide to Dep

### LINKED ISSUE #149: Migrate from Glide to Dep
Glide is no longer in active development and README suggests users to switch to dep as well:

https://github.com/Masterminds/glide

dennwc:
Resolves #149


mcuadros:
Also, we should upload the `vendor` folder.

Gopkg.toml
```diff
@@ -0,0 +1,82 @@
+# Gopkg.toml example
```
juanjux: This probably can removed.

dennwc:
@mcuadros I would rather not. We have a backup of every dependency on each developer's laptop. If we cannot rely on it, when we can ask Data Retirieval to always fetch dependencies of our projects and archive them.

mcuadros:
It's the default in go, and we do this in all the projects.

This is a server so none need to download any dependecy.

dennwc:
@mcuadros I wouldn't say it's the default in Go, but some large projects use this approach.

As a note, by pushing more duplicate files we are growing a number of duplicates that Gemini is trying to detect ;)

Also, it's not true that you will need no dependecies after this - you will still need to pull ostree and build it, so environment setup is more complex anyway. Asking to `dep ensure` is the easiest part, I would say.

On a practical side, I see only two options - either do not push vendor at all, or push it as a separate repo and link to this one as a submodule ([cockroachdb](https://github.com/cockroachdb/cockroach) is [using](https://github.com/cockroachdb/vendored) this approach), which is a bit painful to update.

mcuadros:
As company standard is being updated, so for this pr I want to this here.

If we are going to change this should be at company level.

It's the default when even the tool you are using has the vendor there. And
almost every single project is doing it.

BTW having submodules is the worst idea ever.

dennwc:
@mcuadros I wanted to verify this, so I took 3 large Go projects that I know (k8s, etcd, cockroach) and run [this](https://gist.github.com/dennwc/29c9d2e3637488e13da8c614c9a93fb2) simple program to discover all dependencies of these projects and test if these dependencies contain vendor directory in the repository tree.

Results:

| Total  | Vendors | Clean |
| --- | --- | --- |
| 259 | 24 | 235 |
| | 9.2% | 90.7% |

Feel free to modify the program and update the stats. Or we could always run a more complex analysis, accouting for repository size, number of starts, etc.

But from current results, it's clearly seen that it's not really a default practice to push vendor into repository.

juanjux:
Since this is a matter that should be decided at the company level and not for a single repository it would be better to move this discussion to the company Slack.

mcuadros:
This is not a fair comparison because the dependencies are usually
libraries, and we are speaking here about programs, that none is depending
on this.

In the other hand if you go to, k8s, terraform, dep, docker, and many many
other applications and they are using vendoring.

We don't use vendoring at libraries, we use gopkg.in in this case.

>

dennwc:
@mcuadros Are we talking about bblfshd only or about applying this company-wide? If only for bblfshd, then it's one things, as you mentioned it's actually a server and is not meant to be a library. But I don't want to use this approach for drivers and libraries that we have (sdk, for example).

mcuadros:
This approach is for services/programs, the SDK is a library and the driver
are delivered as docker images.

For libraries we use always gopkg.in

mcuadros:
Test are failing @dennwc

dennwc:
@mcuadros Fixed. It was a problem with wrong gitignore directive.

PR #153, opened 2018-03-03
Run containers in rootless mode

dennwc:
Resolves https://github.com/bblfsh/sdk/issues/225


mcuadros:
Any update about this?

dennwc:
@mcuadros Not yet, sorry. It works for me on the host without root but fails from inside the container. Still need to figure out what permission bits are missing and how to set them properly.

eiso:
Just dropping in to say that it would be exciting if this worked one day. Would save a lot of hassle for people who are in more restrictive environments.

dennwc:
Closing this since we decided to move away from `libcontainer`.

PR #158, opened 2018-03-13
bblfshctl: fix install commands doc and error description

dennwc:


cli/bblfshctl/cmd/driver_install.go
```diff
@@ -133,7 +133,7 @@ func (c *DriverInstallCommand) Execute(args []string) error {
 
 func (c *DriverInstallCommand) Validate() error {
 	if !c.All && !c.Recommended && (c.Args.Language == "" || c.Args.ImageReference == "") {
-		return fmt.Errorf("error --language and --image are mandatory")
+		return fmt.Errorf("error `language` and `image` arguments are mandatory")
```
juanjux: Maybe "positional arguments" instead of "arguments" to make it clearer what they are?

dennwc: Good point! Will update shortly.

PR #170, opened 2018-07-11
Fix flags handling in bblfshctl

dennwc:
Recently I discovered the bug where custom flags for `bblfshctl driver xxxx` won't have any effect.

It turned out that we use go-flags in a wrong way.

We had a `DriverCommand` struct and `DriverListCommand` subcommand that extends it:
```golang
type DriverListCommand struct {
	DriverCommand
}
```

But both structs were registered separately:

```golang
c, _ := parser.AddCommand("driver",
	..., &cmd.DriverCommand{},
)

c.AddCommand("list",
	..., &cmd.DriverListCommand{},
)
```

Now, if we run `bblfshctl driver -x list` the `-x` flag will be applied to a `&cmd.DriverCommand{}` pointer, and `&cmd.DriverListCommand{}` will get a default value of `-x` (flag has no effect). Since the control will be passed to `DriverListCommand` and the pointer of `DriverCommand` will be lost, the command will not see the flag.

An obvious solution is to embed the pointer to the parent command:
```golang
type DriverListCommand struct {
	*DriverCommand
}

// ...
d := &cmd.DriverCommand{}
c, _ := parser.AddCommand("driver",
	..., d,
)

c.AddCommand("list",
	..., &cmd.DriverListCommand{d},
)
```

But this won't work either, since now `DriverListCommand` will change all fields of `d` pointer to default values (since they are still passed to the parent command).

This PR solves the problem by removing any flags from the parent command (`bblfshctl driver`) and leaving only flags for `bblfshctl driver list`.


mcuadros:
You should register only one of the structs, doesn't make sense register the parent.

dennwc:
@mcuadros Sure, but this was in bblfshctl code. The PR fixes it.

PR #174, opened 2018-07-31
bblfshctl: add driver remove --all

campoy:
`bblfshctl driver remove --all` following the same convention as `bblfshctl driver install --all`.

Signed-off-by: Francesc Campoy <campoy@golang.org>

### DIFF
--- a/cmd/bblfshctl/cmd/driver_remove.go
+++ b/cmd/bblfshctl/cmd/driver_remove.go
@@ -9,7 +9,7 @@ import (
 )
 
 const (
-	DriverRemoveCommandDescription = "Removes a new driver for the given language"
+	DriverRemoveCommandDescription = "Removes the driver for the specified language"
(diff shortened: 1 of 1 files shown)


cmd/bblfshctl/cmd/driver_remove.go
```diff
@@ -18,25 +18,46 @@ type DriverRemoveCommand struct {
...
-	if err == nil && len(r.Errors) == 0 {
-		return nil
+	langs := []string{c.Args.Language}
+	if c.All {
+		r, err := c.srv.DriverStates(ctx, &protocol.DriverStatesRequest{})
+		if err != nil {
+			return err
+		}
+
+		if err != nil || len(r.Errors) > 0 {
```
juanjux: Duplicated check of err (there is another above).

campoy: good catch, fixed

PR #185, opened 2018-09-18
Delete existing driver on update

meyskens:
If there is an existing driver and --update is set it should delete the
driver for that language then install the new one. If it doesn't do this
multiple drivers for one language could be installed.

I made this PR based on some weird behavior that I noticed when installing a new driver image with a different tag using `--update`. I am not sure if it is expected to have multiple drivers installed for one language (e.g. the remove only removes by language not image)


daemon/daemon.go
```diff
@@ -74,13 +74,15 @@ func (d *Daemon) InstallDriver(language string, image string, update bool) error
...
-			return ErrAlreadyInstalled.New(language, s.Name())
-		case !ErrMissingDriver.Is(err):
-			return ErrRuntime.Wrap(err)
+
+	s, err := d.getDriverImage(language)
+	if !ErrMissingDriver.Is(err) && !update {
+		return ErrRuntime.Wrap(err)
+	}
+	if err == nil && update {
+		err = d.runtime.RemoveDriver(s)
```
dennwc: If `s` is a driver image ID (not a name), then delete should happen after the successful install of a new driver.
But this will lead to the same weird behavior if delete fails. So please add a `TODO` that this should happen after the install and we will take care of it later.

meyskens: Thanks! will add this!

PR #188, opened 2018-09-25
Build a Docker image with preinstalled drivers

dennwc:
* Update Dockerfile to use multistage build (`docker build .` just works)
* Include a new Makefile target to build/push an image with preinstalled drivers
* Few minor fixes along the way


campoy:
Is there a way I can test the generated Docker image, @dennwc ?

dennwc:
@campoy Sure! There are two ways: using Makefile or Dockerfile.

### Makefile:
```
make build-drivers
```
It will generate an image with a tag similar to `bblfsh/bblfshd:aaa-xxxx-dirty-drivers`. Since it's not possible to control the tag of resulting image I prefer the second way.

### Dockerfile:
```
docker build -t bblfsh/bblfshd:latest .
docker build -f Dockerfile.drivers --build-arg TAG=latest -t bblfsh/bblfshd:latest-drivers .
```
It will generate an image with tag `bblfsh/bblfshd:latest-drivers` (can be customized).

PR #210, opened 2018-11-08
Deprecate v1 parsing protocol

dennwc:
This PR will track a deprecation of v1 parsing protocol.

Note, that gRPC v1 protocol is still used to get a list of all languages and driver version info.

Support in clients:
- [x] Go (https://github.com/bblfsh/go-client/pull/86)
- [x] Python (https://github.com/bblfsh/client-python/pull/128)
- [x] Scala (https://github.com/bblfsh/client-scala/pull/76)

---
This change is [Reviewable](https://reviewable.io/reviews/bblfsh/bblfshd/210)

### DIFF
--- a/daemon/common_test.go
+++ b/daemon/common_test.go
@@ -1,22 +1,25 @@
 package daemon
 
 import (
+	"bytes"
 	"encoding/hex"
 	"fmt"
 	"os"
 	"path/filepath"
 
+	"github.com/containers/image/types"
+	xcontext "golang.org/x/net/context"
+	"google.golang.org/grpc"
+
 	"github.com/bblfsh/bblfshd/daemon/protocol"
 	"github.com/bblfsh/bblfshd/runtime"
-
-	"github.com/containers/image/types"
-	oldctx "golang.org/x/net/context"
-	"google.golang.org/grpc"
 	"gopkg.in/bblfsh/sdk.v1/manifest"
 	protocol1 "gopkg.in/bblfsh/sdk.v1/protocol"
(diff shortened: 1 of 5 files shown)


bzz:
Quick question - as this is incompatible API changes would this, according to SemVer, mean that after this is merged we are going to release `bblfshd` v3.0.0 ?

That would follow the logic and match the behavior of client-go.v3 version.

daemon/common_test.go
```diff
@@ -1,22 +1,25 @@
...
 	"fmt"
 	"os"
 	"path/filepath"
 
-	"github.com/bblfsh/bblfshd/daemon/protocol"
-	"github.com/bblfsh/bblfshd/runtime"
-
 	"github.com/containers/image/types"
-	oldctx "golang.org/x/net/context"
+	xcontext "golang.org/x/net/context"
```
creachadair: Do we still support any Go releases old enough that this couldn't be more simply `"context"` in the standard section above?

daemon/driver.go
```diff
@@ -22,13 +22,20 @@ import (
 	protocol2 "gopkg.in/bblfsh/sdk.v2/protocol"
 )
 
+type ServiceV1 interface {
```
creachadair: Please document all the exported members. In this case there's a bunch of existing use, please add comments at least for the new names introduced here (cf. https://github.com/golang/go/wiki/CodeReviewComments#doc-comments).

I noticed a lot of our files lack clear doc comments. I'll file an issue for adding them in a more comprehensive way to the existing code.

kuba--:
Closing for now

Issue #211, opened 2018-11-19
Ensure exported names in Go packages are properly documented


creachadair:
There are a number of packages missing conventional documentation comments (cf. https://github.com/golang/go/wiki/CodeReviewComments#doc-comments). Given that we advertise these libraries as part of an SDK, we should make an effort to back-fill the ones that are missing.

This helps not only us (speaking as the new person trying to learn his way around), but also our users, who should be able to use godoc.org to good effect.

bzz:
👍 and we could also enable `go lint` with [this check](https://github.com/golang/lint/blob/master/lint.go#L476) on CI, to avoid further API doc gaps.

creachadair:
The `golint` checks are mostly good, but there are enough false positives that I'd be reluctant to block changes for all of them. That said, I agree it'd be nice to have some pre-submit linting suggestions.

dennwc:
We may disable most `golint` checks and leave only ones related to documentation, for example.

PR #213, opened 2018-11-20
Graceful report for non-existing image

### LINKED ISSUE #207: Trying to install a non-existent driver gives authorization error
From https://github.com/src-d/engine/issues/104 (thanks @BFergerson):

> I accidentally typed the command `srcd parse drivers install list` instead of `srcd parse drivers list` and received error:

```
INFO[0000] installing driver                             language=list version=latest
ERRO[0000] unable to install version latest of list driver: rpc error: code = Unknown desc = can't install list driver: rpc error: code = Unknown desc = errors:
denied: requested access to the resource is denied
unauthorized: authentication required
```

> `authentication required` is an odd error for this mistaken command.

bzz:
Similar to `AlreadyInstalled` status from d5b686f8168286df965c8b3a6950d132a0794d87 this change propagates the fact that [image request is not found or needs to have authentication credentials](https://github.com/docker/distribution/blob/master/registry/client/errors.go) though gRPC `Unauthenticated` code on the `driver install` command. 

Before:
```
$ bblfshctl driver install list12223
Installing driver from "list12223"... rpc error: code = Unknown desc = errors:
denied: requested access to the resource is denied
unauthorized: authentication required
```

After:
```
$ bblfshctl driver install list12223
Installing driver from "list12223"... driver does not exist or is private
```

It resolves #207 in a way it's been decided in that issue.

While at it, few other small fixes included:
 - [x] latest `dep ensure`
 - [x] `bblfshctl -ctl-endpoint` -> `bblfshctl -ctl-network` for consistency, same as it is in `bblfshd`
 - [x] tiny err msg cleanup


cmd/bblfshctl/cmd/driver_install.go
```diff
@@ -299,6 +299,8 @@ func (c *DriverInstallCommand) installDriver(ctx context.Context, ref driverRef)
 	})
 	if st, ok := status.FromError(err); ok && st.Code() == codes.AlreadyExists {
 		return daemon.ErrAlreadyInstalled.New(ref.Lang, ref.Ref)
+	} else if st, ok := status.FromError(err); ok && st.Code() == codes.Unauthenticated {
```
dennwc:
```suggestion
	} else if ok && st.Code() == codes.Unauthenticated {
```

daemon/daemon_test.go
```diff
@@ -42,6 +43,16 @@ func TestDaemonInstallDriver(t *testing.T) {
 	require.Nil(err)
 }
 
+func TestDaemon_InstallNonexistentDriver(t *testing.T) {
+	require := require.New(t)
+	s, tmp := buildMockedDaemon(t)
+	defer os.RemoveAll(tmp)
+
+	err := s.InstallDriver("", "docker://list", false)
+	require.Error(err, "An error was expected")
```
dennwc:
```suggestion
	require.NotNil(err, "An error was expected")
```

daemon/protocol/service.go
```diff
@@ -121,6 +121,12 @@ func (s *protocolServiceServer) InstallDriver(ctx xcontext.Context, req *Install
 
 	if ErrAlreadyInstalled.Is(err) {
 		return nil, status.New(codes.AlreadyExists, err.Error()).Err()
+	} else if errs, ok := err.(errcode.Errors); ok { //docker err codes analysis
+		for _, erro := range errs {
+			if strings.HasPrefix(erro.Error(), "unauthorized:") {
```
dennwc: This should type-assert an error instead of checking the message text

daemon/daemon_test.go
bzz: Could you please elaborate on this suggestion, to help understand it better? 

From [require.Error GoDoc](https://godoc.org/github.com/stretchr/testify/require#Error)
```
Error asserts that a function returned an error (i.e. not `nil`).
```

daemon/protocol/service.go
bzz: That would be lovely, but from the experiments that I did - it does not seem possible with their usage of [errcode.Errors](https://github.com/docker/distribution/blob/master/registry/api/errcode/errors.go#L176): al `reflect.TypeOf(erro)` returns the same type.

Or am I missing something here?

creachadair: This is the trouble with implementing `error` on a raw slice. You could probably give it a method to walk through the slice and pull out the first coded value—but in the end the problem is that we're not using an error as a value but as a container. 

It'd be nice to have a test for this, though.

daemon/daemon_test.go
dennwc: Sorry, confused it with a similar function from a different package :)

daemon/protocol/service.go
dennwc: @bzz Getting error code is what I assumed. I think it will be better than relying on a specific error prefix, since the error codes should be stable.

daemon/protocol/service_test.go
```diff
@@ -0,0 +1,69 @@
...
+)
+
+type mockedServiceAnyOtherErr struct {
+	mock.Mock
+	Service
+}
+
+func (s *mockedServiceAnyOtherErr) InstallDriver(language string, image string, update bool) error {
+	var errors errcode.Errors = []error{errors.New("any other error")}
+	return errors
```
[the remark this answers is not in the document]
bzz: Added in ee86a03

daemon/errors.go
```diff
@@ -6,9 +6,13 @@ import (
 )
 
 var (
-	ErrUnexpected        = errors.NewKind("unexpected error")
-	ErrMissingDriver     = errors.NewKind("missing driver for language %q")
-	ErrRuntime           = errors.NewKind("runtime failure")
-	ErrAlreadyInstalled  = protocol.ErrAlreadyInstalled
+	ErrUnexpected       = errors.NewKind("unexpected error")
```
creachadair: Please document all these errors (so they have something explanatory in godoc).

bzz: Have documented the only one error so far, that was added in this PR.

The rest, to the best of my knowlege, is added in e3f38d8

Would prefer to do further godoc updates for the code that was not changed in PR in a separate effort like #211

PR #216, opened 2018-11-21
Refactoring: use gRPC server config logic from SDK

bzz:
Part of the bblfsh/client-go#102, migrates to https://github.com/bblfsh/sdk/pull/329

TODOs:
 - [x] refactor bblfshd server to use server gRPC configuration from https://github.com/bblfsh/sdk/pull/329
 - [x] update SDK in vendored deps of `bblfshd` ~(as soon as as soon as https://github.com/bblfsh/sdk/pull/329 merged and new SDK version is released)~ [v2.8.0](https://github.com/bblfsh/sdk/releases/tag/v2.8.0) 🚀 
 - [ ] fix CI


bzz:
Changed to WIP as I'm looking into CI failure.

bzz:
Tested and ready for review. CI is expected to be red until all TODO from PR description are not finished.

cmd/bblfshd/main.go
```diff
@@ -94,7 +94,12 @@ func main() {
 	logrus.Infof("bblfshd version: %s (build: %s)", version, build)
 
 	r := buildRuntime()
-	d := daemon.NewDaemon(version, r, buildGRPCOptions()...)
+	grpcOpts, err := sdk.BuildGRPCOptions(*maxMessageSize)
+	if err != nil {
+		logrus.Errorln(err)
```
juanjux: Shouldn't this also exit the function? (you're using grpcOpts later even if `err != nil`)

cmd/bblfshd/main.go
```diff
@@ -94,7 +94,12 @@ func main() {
 	logrus.Infof("bblfshd version: %s (build: %s)", version, build)
 
 	r := buildRuntime()
-	d := daemon.NewDaemon(version, r, buildGRPCOptions()...)
+	grpcOpts, err := cmdutil.GRPCSizeOptions(*maxMessageSize)
+	if err != nil {
+		logrus.Warnln(err)
```
dennwc: I also think this is the right place to fail and exit the program.

bzz: @juanjux @dennwc but what is the advantage of doing that?

In current ([documented](https://github.com/bblfsh/sdk/pull/329/files#diff-3d26f06e2a3da8bfb583a3f488a3baafR28)) behavior if this happens, we just cap the limit to 100MiB, report it to the user as warning and continue running.

I belive this has the advantage of much better user experience - it does the right thing recovering automatically from this simple error, falling back to reasonable defaults and letting user know about that.

I'm open to arguments of benefits of previous approach though!

cmd/bblfshd/main.go
bzz: If that is allright, let's keep this discussion in https://github.com/bblfsh/bblfshd/pull/216#discussion_r235770197 and it would be good to have arguments there, why it should. I tried listing once that I see why it should not in there :)

cmd/bblfshd/main.go
bzz: To address arguments from https://github.com/bblfsh/sdk/pull/329#discussion_r235762041

@dennwc 
> This is an error on the program startup

yes, but it is a recoverable one - so we recover. I'm trying to understand what is the benefit of moving cognitive load of doing that to the user, if software can do it with a reasonable default? So far, I can not see

@juanjux @creachadair 
> it's using grpcOpts obtained from a call that returned an error, it definitely should exit the program or at least the function.

Why?  For me "Idiomatic code" <<<<< "User Experience" and plus, as current behavior is not idiomatic, it's well documented.

creachadair: We should not force ourselves to choose between user experience and idiomatic code. The reason we have this difficult now, as far as I can tell, is that we have used error-handing to propagate what should have been a diagnostic log.

My view is we should not try to resolve this problem here and now, but take a more systematic approach to separating places where we've used errors to indicate "unfortunate" rather than "erroneous" conditions. Then we can go back and clean these things up. A proper solution will probably (as @bzz pointed out) require some changes to logger propagation.

creachadair: How often do we expect to hit this limit in practice? Is it often enough that we should consider adding chunking (on the request side) and pagination (on the response)? 

In other words: If it's really important to the caller to set a specific value here (e.g., because they have huge files and can't get work done), should we make the API support that without requiring the user to think about gRPC internals? 

This again is a larger question than we can solve right here in this PR, but I'm trying to get an understanding of the decision space. Can we avoid forcing the user to think about this at all?

dennwc: Hardcoding the limit and adding a chunking (streaming) endpoint sounds like the best option to me. As a separate PR, of course :)

bzz: @creachadair 
> We should not force ourselves to choose between user experience and idiomatic code. 

totally agree.

> used error-handing to propagate what should have been a diagnostic log

thank you for ☝️, yes, that is the case now.

@dennwc 
> I don't think it's a recoverable error.
> Let's say the default .... The user tries to sends ..

Thank you for illustrative example @dennwc, it really aids understand and makes 👍argument.

>  As a note, the issue is a bit worse, since the log message will pop up in the log of the driver, not in bblfshd

A bit confused how logs from [`cmd/bblfshd/main.go`](https://github.com/bblfsh/bblfshd/pull/216/files#diff-94bf84897cc9a7746dc1bac15c4484eaR97) can pop up in a driver. Driver gRPC server process has it's own logs, where similar condition was also taken care of. 

Anyway, thanks everyone for productive discussion - I think https://github.com/bblfsh/bblfshd/pull/216#discussion_r235792138 is an overriding argument here, I will change it back to error condition.

On
>> How often do we expect to hit this limit in practice? Is it often enough that we should consider adding chunking (on the request side) and pagination (on the response)?
>
> Hardcoding the limit and adding a chunking (streaming) endpoint sounds like the best option

I think this is going to be quite often, as soon as gitbase is used to process a substantial fraction of Github.

The fact that nobody of the downstream clients in DR has stumbled up on this limitation most probably means that they work it around by only parsing a lines smaller then the limit (a [quick search](https://github.com/src-d/gitbase/blob/master/internal/function/uast.go#L28) verifies this guess)

Having streaming gRPC endpoint sounds like possible solution, but before implementing it I would prefer to have a separate issue \w a proposal and then verify @creachadair concerns with more clients/users first.

WDYT?

bzz:
Review feedback addressed in 4ae7a3893bf48ed8f84b4b5acef1d68061acadc0, ready for another round.

cmd/bblfshd/main.go
creachadair: I think for now we should set the limits high enough to push the tail below some limit of interest. That has the advantage of being something we can do now to unblock other work. By the time you hit 2GiB on the wire, it's perfectly reasonable to report to the user "you're doing something weird" and let them come talk to us.

I agree that a streaming solution is a bigger project and deserves a separate design process.

bzz: > I think for now we should set the limits high enough

I belive that is exactly what we have done here - it is set 100MiB for now.

bzz:
2 tests are failing on CI 🤔 

```
--- FAIL: TestDaemonInstallDriver (7.91s)
	Error Trace:	daemon_test.go:42
	Error:		Expected nil, but got: Invalid status code returned when fetching blob 404
		
--- FAIL: TestDriverInstance_State (6.33s)
	Error Trace:	driver_test.go:95
			driver_test.go:47
	Error:		Received unexpected error Invalid status code returned when fetching blob 404
			github.com/bblfsh/bblfshd/vendor/github.com/containers/image/docker.(*dockerImageSource).GetBlob
				/go/src/github.com/bblfsh/bblfshd/vendor/github.com/containers/image/docker/docker_image_src.go:173
			github.com/bblfsh/bblfshd/vendor/github.com/containers/image/image.(*manifestSchema2).ConfigBlob
				/go/src/github.com/bblfsh/bblfshd/vendor/github.com/containers/image/image/docker_schema2.go:107
			github.com/bblfsh/bblfshd/vendor/github.com/containers/image/image.(*manifestSchema2).OCIConfig
				/go/src/github.com/bblfsh/bblfshd/vendor/github.com/containers/image/image/docker_schema2.go:86
			github.com/bblfsh/bblfshd/vendor/github.com/containers/image/image.(*sourcedImage).OCIConfig
				<autogenerated>:1
			github.com/bblfsh/bblfshd/runtime.(*driverImage).WriteTo
				/go/src/github.com/bblfsh/bblfshd/runtime/image.go:80
			github.com/bblfsh/bblfshd/runtime.(*storage).Install
				/go/src/github.com/bblfsh/bblfshd/runtime/storage.go:61
			github.com/bblfsh/bblfshd/runtime.(*Runtime).InstallDriver
				/go/src/github.com/bblfsh/bblfshd/runtime/runtime.go:62
			github.com/bblfsh/bblfshd/daemon.NewRuntime
				/go/src/github.com/bblfsh/bblfshd/daemon/driver_test.go:94
			github.com/bblfsh/bblfshd/daemon.TestDriverInstance_State
				/go/src/github.com/bblfsh/bblfshd/daemon/driver_test.go:47
			testing.tRunner
				/usr/local/go/src/testing/testing.go:777
			runtime.goexit
				/usr/local/go/src/runtime/asm_amd64.s:2361
```

juanjux:
Since CI seems to be working I'm merging.

PR #217, opened 2018-11-22
bblfshctl: fix NPE in driver remove

### LINKED ISSUE #214: bblfshctl: panic on empty args driver remove
Empty arguments for `bblfshctl driver remove` causes process to panic.

```
$ bblfshctl driver remove

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0xbdbaca]
goroutine 1 [running]:
github.com/bblfsh/bblfshd/cmd/bblfshctl/cmd.(*DriverRemoveCommand).Execute(0xc0000afe00, 0xc000253660, 0x0, 0x2, 0xc0000afe00, 0x1)
        /gopath/src/github.com/bblfsh/bblfshd/cmd/bblfshctl/cmd/driver_remove.go:52 +0x1da
github.com/bblfsh/bblfshd/vendor/github.com/jessevdk/go-flags.(*Parser).ParseArgs(0xc000ab2780, 0xc0000300a0, 0x2, 0x2, 0xc000253540, 0x2, 0x4, 0xc0001746c0, 0xc0004d9480
)
        /gopath/src/github.com/bblfsh/bblfshd/vendor/github.com/jessevdk/go-flags/parser.go:316 +0x800
github.com/bblfsh/bblfshd/vendor/github.com/jessevdk/go-flags.(*Parser).Parse(0xc000ab2780, 0xe3a936, 0x6, 0xf2ffeb, 0x2d, 0xf2ffeb)
        /gopath/src/github.com/bblfsh/bblfshd/vendor/github.com/jessevdk/go-flags/parser.go:186 +0x71
main.main()
        /gopath/src/github.com/bblfsh/bblfshd/cmd/bblfshctl/main.go:57 +0x434
```

Expected result: usage printed.

dennwc:
Fixes #214

### DIFF
--- a/cmd/bblfshctl/cmd/driver_remove.go
+++ b/cmd/bblfshctl/cmd/driver_remove.go
@@ -48,12 +48,13 @@ func (c *DriverRemoveCommand) Execute(args []string) error {
 
 	for _, lang := range langs {
 		r, err := c.srv.RemoveDriver(ctx, &protocol.RemoveDriverRequest{Language: lang})
-		if err != nil || len(r.Errors) > 0 {
+		if err != nil {
+			return err
+		} else if len(r.Errors) != 0 {
 			for _, e := range r.Errors {
 				fmt.Fprintf(os.Stderr, "Error, %s\n", e)
 			}
-
-			return err
+			return fmt.Errorf("driver remove failed: %v", r.Errors)
 		}
 	}
 	return nil


cmd/bblfshctl/cmd/driver_remove.go
```diff
@@ -48,12 +48,13 @@ func (c *DriverRemoveCommand) Execute(args []string) error {
 
 	for _, lang := range langs {
 		r, err := c.srv.RemoveDriver(ctx, &protocol.RemoveDriverRequest{Language: lang})
-		if err != nil || len(r.Errors) > 0 {
+		if err != nil {
+			return err
```
juanjux: Alternative: check for `r != nil` before this block and print the errors, then exit (or not, if `err == nil`). This should cover the case where `err != nil` and `r` contains informative errors (which I don't know if it's the case with the current implementation of RemoveDriver, but could be in the future).

dennwc: `RemoveDriver` is from the auto-generated client, so it won't ever return both non-nil error and non-nil response.

bzz: I wonder if there is a way to express this kind of invariant in the code 🤔 

As I do not see a way for language to support this, obvious one would be to have a test..

PR #225, opened 2018-11-27
Integrate opentracing

dennwc:
Enable OpenTracing and instrument some operations related to driver management in `bblfshd`.

Also, pass relevant environment variables to language drivers, so the tracing works for the whole service stack.

Blocked by https://github.com/bblfsh/sdk/pull/330


daemon/daemon.go
```diff
@@ -19,22 +24,45 @@ type Daemon struct {
...
 
-	version string
-	runtime *runtime.Runtime
-	pool    map[string]*DriverPool
-	mutex   sync.Mutex
+	version   string
+	runtime   *runtime.Runtime
+	driverEnv []string
+	pool      map[string]*DriverPool
+	mutex     sync.Mutex
```
creachadair: Conventionally, a mutex goes before the fields it protects. If that's not the convention we're using here, please include a comment to say which fields it protects.

Separately: I recommend calling this more simply `mu` (which is a fairly common convention for a mutex field).

bzz:
CI seems to be failing

daemon/daemon.go
dennwc: Oh, sure, I haven't yet refactored it. Both comments make perfect sense.

bzz:
@dennwc sorry if I have missed it elsewhere, but do you think we should also very briefly document this?

I suggest adding a new section consisting of single sentence to the https://github.com/bblfsh/bblfshd#development on how to enable this instrumentation.

dennwc:
@bzz Good catch! I will include a section in the docs.

dennwc:
@bzz Added short documentation, PTAL.

PR #228, opened 2018-12-12
Use client's context and don't force max timeout

### LINKED ISSUE #227: Unable to allocate a driver instance: timeout, all drivers are busy
On gitbase we are having the following error if we try to parse more than 1000 files:

`bblfshd     | time="2018-11-30T13:13:35Z" level=warning msg="unable to allocate a driver instance: timeout, all drivers are busy" language=go`

https://github.com/src-d/gitbase/issues/634

Even changing the request timeout we still have the error.

Is this timeout a different one? If yes, is it possible to change from the client side?

dennwc:
Currently `bblfshd` forces a timeout on the context to be 5 min at most. When doing batch processing, the client may send thousands of requests, saturating the pool. Depending on the size of source files, clients may hit a max timeout and they have no way to wait longer, even if they are willing to.

This change drops the maximal timeout and relies on the timeout provided by the client (via context). The deadline propagation works over gRPC, allowing clients to specify exactly what time they are willing to wait for the response.

If the deadline is not set, `bblfshd` will set it to a default value automatically, as before.

Fixes #227.

### DIFF
--- a/daemon/pool.go
+++ b/daemon/pool.go
@@ -4,7 +4,6 @@ package daemon
 
 import (
 	"context"
-	"fmt"
 	"math"
 	"runtime"
 	"sync"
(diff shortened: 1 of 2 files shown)


juanjux:
CI is failing.

PR #233, opened 2019-01-28
Fix build time on VersionResponse

juanjux:
Since it wasn't being loaded on `NewDaemon` on main (or elsewhere) the nice date 0001-01-01 was being sent by RPC.

Related to: bblfsh/client-python/issues/141

Signed-off-by: Juanjo Alvarez <juanjo@sourced.tech>

### DIFF
--- a/daemon/daemon.go
+++ b/daemon/daemon.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"strings"
 	"sync"
+	"time"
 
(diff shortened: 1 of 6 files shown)


daemon/daemon.go
```diff
@@ -32,12 +34,13 @@ type Daemon struct {
 }
 
 // NewDaemon creates a new server based on the runtime with the given version.
```
creachadair: Please also update the comment to describe what the `build` argument means.

PR #239, opened 2019-01-31
Always honor user-provided timeout

dennwc:
Right now we are forcing a timeout of 5 sec even if the user wants to wait for any time for a request to complete.

This change removes the defaults for v2 gRPC API. Disabling it in v1 requires more work and will be done in a separate PR.


daemon/pool.go
```diff
@@ -18,10 +18,12 @@ import (
...
 )
 
-const (
+var (
 	// DefaultPoolTimeout is the time a request to the DriverPool can wait
 	// before getting a driver assigned.
-	DefaultPoolTimeout = 5 * time.Second
+	//
+	// Deprecated: use the context to specify the timeout; default was set to 5 sec
+	DefaultPoolTimeout struct{}
```
creachadair: Can we get rid of this entirely? By changing the type we've already made a breaking change in the package API, so at that point we might as well just drop the name entirely I think.

dennwc: Just wanted to have a piece of docs there to specify the solution. My first thought was to drop it as well, so if it's not an issue, I'll do it.

creachadair: I don't object to dropping it. It would be worth having a note in the release notes for the next SDK release, though.

PR #241, opened 2019-01-31
Specify an exact date format for the Makefile

dennwc:
It looks like `date -Iseconds` is not portable - it gives different results locally and in CI.

This PR specifies an exact format that we expect.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,7 +25,7 @@ BUILD_PATH := $(BASE_PATH)/build
 CMD_PATH := $(BASE_PATH)/cmd
 
 # Build information
-BUILD ?= $(shell date -Iseconds)
+BUILD ?= $(shell date +%FT%X%z)
 GIT_COMMIT=$(shell git rev-parse HEAD | cut -c1-7)
 GIT_DIRTY=$(shell test -n "`git status --porcelain`" && echo "-dirty" || true)
 DEV_PREFIX := dev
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -25,7 +25,7 @@ BUILD_PATH := $(BASE_PATH)/build
 CMD_PATH := $(BASE_PATH)/cmd
 
 # Build information
-BUILD ?= $(shell date -Iseconds)
+BUILD ?= $(shell date +%FT%X%:z)
```
creachadair: Note that this format doesn't work with macOS:

```shell
$ date +%FT%X%:z
2019-01-31T10:50:56:z
```

dennwc: Is there something else that can produce `2019-01-31T20:36:34+02:00`?

I feel like I'll write a tiny Go program that will print the date in the right format.

creachadair: That seems like overkill.

Just leaving out the extra colon should get you most of the way there:

```
$ date +%FT%X%z
2019-01-31T10:59:58-0800
```

but note it omits the colons between the hours/minutes of the Z offset.

creachadair: It looks like the Linux `date` command supports `%z` without the colons for a similar effect, so if we don't care about the exact format but only consistency, that should probably work.

dennwc: Yeah, sorry for the delay, was working on the other issue. Pushed the `%z` and changed the parse format.

PR #242, opened 2019-02-01
Quote linker flags passed to the Go build

creachadair:
Because the "go build" invocation strips one layer of quoting, we need to ensure that any extra spaces in the argument list do not trigger spurious extra arguments to the underlying go tool link invocation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,7 +53,7 @@ ifneq ($(origin TRAVIS_TAG), undefined)
 endif
 
 # Build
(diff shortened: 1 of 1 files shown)


creachadair:
I'm going to go ahead and merge this to unblock Maxim.

creachadair:
This turns out not to have been enough. But I missed it because the validation build for the PR worked OK, it didn't fail till deployment. I am still working to isolate the reason for that.

juanjux:
Yeah, issues involving CI usually mean lots of PRs and trial+error.

Issue #243, opened 2019-02-01
Deployment builds generate an invalid Go linker command-line


creachadair:
When Travis builds for deployment, e.g., upon seeing a new release tag, it generates a build stamp with a space in it. This unfortunately breaks when the `go` tool strips off the outside quotes from its `-ldflags` argument, because it results in an extra unrecognized argument.

An example from a recent push build:

```
BBLFSHD_BUILD="2019-02-01T05:51:47 PM+0000"
```

This is not the time format we wanted, which is (for example):

```shell
$ date +%FT%X%z
2019-02-01T10:07:41-0800
```

I'm not yet sure what differs in the deployment config.

There are, I think, two things to address here:

- [x] The date string is coming out differently in the release build than the test build (#246).
- [x] The version string quoting is being lost (#244).

creachadair:
Ok! Progress: The quoting issues are _actually_ fixed now. The CI still [generates a different timestamp](https://travis-ci.org/bblfsh/bblfshd/builds/489161481), e.g. `2019-02-05T05:54:55 PM+0000` instead of `2019-02-05T05:54:55+0700` (note the inserted `PM`).

creachadair:
This is now fixed and deployed in release v2.11.4.

PR #244, opened 2019-02-01
Quote linker flags passed to go build

creachadair:
Updates #243. I attempted to fix this in #242, but I put the quotes in the wrong place: The offending node is in the Dockerfile for the CD push.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,10 +22,10 @@ RUN mkdir /build
 ARG BBLFSHD_VERSION=dev
 ARG BBLFSHD_BUILD=unknown
 
-ENV GO_LDFLAGS="-X main.version=${BBLFSHD_VERSION} -X main.build=${BBLFSHD_BUILD}"
+ENV GO_LDFLAGS="-X main.version=${BBLFSHD_VERSION} -X main.build='${BBLFSHD_BUILD}'"
 
-RUN go build  -tags ostree --ldflags "${GO_LDFLAGS}" -o /build/bblfshd ./cmd/bblfshd/
-RUN go build --ldflags "${GO_LDFLAGS}" -o /build/bblfshctl ./cmd/bblfshctl/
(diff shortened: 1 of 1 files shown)


creachadair:
I am able to reproduce the CD failure by locally bugging the date format to inject an extra space. Fortunately, the `-ldflags` parsing understands quoting, so we can protect ourselves from that.

creachadair:
N.B.: Although #242 was not the right fix for this issue, it was actually not a bad change. It was just not necessary.

Dockerfile
```diff
@@ -22,10 +22,10 @@ RUN mkdir /build
 ARG BBLFSHD_VERSION=dev
 ARG BBLFSHD_BUILD=unknown
 
-ENV GO_LDFLAGS="-X main.version=${BBLFSHD_VERSION} -X main.build=${BBLFSHD_BUILD}"
+ENV GO_LDFLAGS="-X main.version=${BBLFSHD_VERSION} -X main.build='${BBLFSHD_BUILD}'"
 
-RUN go build  -tags ostree --ldflags "${GO_LDFLAGS}" -o /build/bblfshd ./cmd/bblfshd/
-RUN go build --ldflags "${GO_LDFLAGS}" -o /build/bblfshctl ./cmd/bblfshctl/
+RUN go build  -tags ostree --ldflags "'${GO_LDFLAGS}'" -o /build/bblfshd ./cmd/bblfshd/
```
bzz: I figured that having only ` --ldflags '${GO_LDFLAGS}'` does not solve the issue, right?

creachadair: Right—the shell won't interpolate into single quotes. By double-quoting the whole word, the single quotes inside lose their meaning until the substitution is complete, and thus survive to protect it from the subsequent evaluation. 

Shell languages are bad and we rely on them entirely too much. :(

PR #248, opened 2019-02-07
HotFix build date parsing

### LINKED ISSUE #247: Binary released as part of v2.11.4 can not parse the build date
Reported in Slack on recent [v2.11.4](https://github.com/bblfsh/bblfshd/releases/tag/v2.11.4).

When running on linux, a release tarball contains a binary that can not be started.

Steps to reproduce:

```
cat /etc/issue
 Debian GNU/Linux 9 \n \l
wget https://github.com/bblfsh/bblfshd/releases/download/v2.11.4/bblfshd_v2.11.4_linux_amd64.tar.gz
tar xvzf bblfshd_v2.11.4_linux_amd64.tar.gz
sudo apt-get install libostree-dev

sudo ./bblfshd_linux_amd64/bblfshd
 [2019-02-07T11:54:34+01:00]  INFO bblfshd version: v2.11.4 (build: "2019-02-05T19:16:31+0000")
 [2019-02-07T11:54:34+01:00]  INFO initializing runtime at /var/lib/bblfshd
 [2019-02-07T11:54:34+01:00] ERROR wrong date format for build: parsing time ""2019-02-05T19:16:31+0000"" as "2006-01-02T15:04:05-0700": cannot parse ""2019-02-05T19
:16:31+0000"" as "2006"
```

The reason seems to be extra `"` in the build time.

Plan 

 - [x] reproduce on local linux build
 - [x] try reverting https://github.com/bblfsh/bblfshd/pull/242/
 - [x] try `strings.Trim(build , ""\")`
 - [x] restore `go build ./cmd/bblfshd` behavior from #233, by making server similar to defaults for the [drivers](https://github.com/bblfsh/bblfshd/blob/49f427820b544393c2891c5d9455f529f784c260/daemon/service.go#L313) in case the date was not provided.

TEST PLAN: 
   - container works
    `make build && docker run -d --name bblfshd-bug --privileged -p 9432:9432 bblfsh/bblfshd:dev-2e8884b-dirty`
   - `bblfshd` binary works
    `make packages && ./build/bblfshd_linux_amd64/bblfshd` 
   - dev build of `bblfshd` bianry works
    `go build ./cmd/bblfshd && ./bblfshd`

bzz:
Fixes #247 (please read it for more context)

Consists of (see individual commits)
 - reverting #242 
 - small refactoring to unify date parsing
 - sync quotation \w `src-d/ci` as a standard
 - restoring `go build` workflow after #233

### DIFF
--- a/cmd/bblfshd/main.go
+++ b/cmd/bblfshd/main.go
@@ -22,9 +22,11 @@ import (
 	"gopkg.in/bblfsh/sdk.v2/driver/manifest/discovery"
 )
 
+const defaultBuild = "undefined"
+
 var (
 	version = "undefined"
-	build   = "undefined"
+	build   = defaultBuild
 
 	network        *string
 	address        *string
@@ -117,8 +119,13 @@ func main() {
 
 	parsedBuild, err := time.Parse(daemon.BuildDateFormat, build)
 	if err != nil {
-		logrus.Errorf("wrong date format for build: %s", err)
-		os.Exit(1)
+		logrus.Errorf("wrong date format for this build: %s", err)
+		if build == defaultBuild {
+			build = time.Now().Format(daemon.BuildDateFormat)
+			logrus.Infof("using start time instead in this dev build: %s", build)
+		} else {
+			os.Exit(1)
+		}
 	}
 	d := daemon.NewDaemon(version, parsedBuild, r, grpcOpts...)
 	if args := cmd.Args(); len(args) == 2 && args[0] == "install" && args[1] == "recommended" {
--- a/cmd/bblfshd/main.go
+++ b/cmd/bblfshd/main.go
@@ -22,7 +22,10 @@ import (
 	"gopkg.in/bblfsh/sdk.v2/driver/manifest/discovery"
 )
 
-const defaultBuild = "undefined"
+const (
+	defaultBuild    = "undefined"
+	buildDateFormat = "2006-01-02T15:04:05-0700"
+)
 
 var (
 	version = "undefined"
@@ -117,12 +120,12 @@ func main() {
 		os.Exit(1)
 	}
 
-	parsedBuild, err := time.Parse(daemon.BuildDateFormat, build)
+	parsedBuild, err := time.Parse(buildDateFormat, build)
 	if err != nil {
 		if build == defaultBuild {
 			parsedBuild = time.Now()
 			logrus.Infof("using start time instead in this dev build: %s",
-				parsedBuild.Format(daemon.BuildDateFormat))
+				parsedBuild.Format(buildDateFormat))
 		} else {
 			logrus.Errorf("wrong date format for this build: %s", err)
 			os.Exit(1)
--- a/cmd/bblfshd/main.go
+++ b/cmd/bblfshd/main.go
@@ -119,11 +119,12 @@ func main() {
 
 	parsedBuild, err := time.Parse(daemon.BuildDateFormat, build)
 	if err != nil {
-		logrus.Errorf("wrong date format for this build: %s", err)
 		if build == defaultBuild {
-			build = time.Now().Format(daemon.BuildDateFormat)
-			logrus.Infof("using start time instead in this dev build: %s", build)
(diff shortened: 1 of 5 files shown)


bzz:
Tested, both Docker and local build on linux work as expected

daemon/daemon.go
```diff
@@ -19,6 +19,8 @@ import (
 	protocol2 "gopkg.in/bblfsh/sdk.v2/protocol"
 )
 
+const BuildDateFormat = "2006-01-02T15:04:05-0700"
```
dennwc: This is not really related to the daemon package. It's only the `bblfshd/main.go` who parses it that way.

cmd/bblfshd/main.go
```diff
@@ -115,10 +117,15 @@ func main() {
...
 	}
 
-	parsedBuild, err := time.Parse("2006-01-02T15:04:05-0700", build)
+	parsedBuild, err := time.Parse(daemon.BuildDateFormat, build)
 	if err != nil {
-		logrus.Errorf("wrong date format for build: %s", err)
-		os.Exit(1)
+		logrus.Errorf("wrong date format for this build: %s", err)
+		if build == defaultBuild {
+			build = time.Now().Format(daemon.BuildDateFormat)
```
dennwc: Why not leave it as `undefined`, but skip an error?

cmd/bblfshd/main.go
```diff
@@ -115,10 +117,15 @@ func main() {
 		os.Exit(1)
 	}
 
-	parsedBuild, err := time.Parse("2006-01-02T15:04:05-0700", build)
+	parsedBuild, err := time.Parse(daemon.BuildDateFormat, build)
 	if err != nil {
-		logrus.Errorf("wrong date format for build: %s", err)
-		os.Exit(1)
+		logrus.Errorf("wrong date format for this build: %s", err)
```
dennwc: This error should be in `else` for `undefined`.

cmd/bblfshd/main.go
bzz: because interface of the NewDaemon has changed and it requires parsed `time.Time` no?

daemon/daemon.go
bzz: It's just retrofitting existing existing code structure to minimize the changes - if you look below, this is used in `daemon_test.go`, thus the location.

cmd/bblfshd/main.go
bzz: I like the idea of this being logged in any case, but can move to the `os.Exit()` branch, if that is what you are asking for.

dennwc: Yes, I think exit is better. Running dev version shouldn't lead to errors like this :)

cmd/bblfshd/main.go
dennwc: Sure, but you still pass parsed time (zero `time.Time`) to `NewDaemon` in this case. It doesn't matter what `build` is.

daemon/daemon.go
dennwc: But the test is free to use any format. `NewDaemon` doesn't enforce any specific format.

bzz: sure, but usually it's kind of good idea to test with the format that is going to be used in prod, no?

cmd/bblfshd/main.go
bzz: Oh, that is a bug, thank you for catching!

bzz: addressed in parsedBuild

daemon/daemon.go
dennwc: The point is that we should either move the constant to this package and dictate the format there (that is not a good idea I think) or somehow write the test for the binary itself (`package main`).

bzz: Got it! Decouple date format parsing and daemon package tests in d4ea40b

bzz:
All feedback addressed, ready for another round @dennwc

bzz:
Merging&releasing a new version as soon as CI is green.

PR #250, opened 2019-02-12
Fix multiple issues with a driver pool

dennwc:
This PR fixes multiple issues found while addressing the timeout issue in C# driver: https://github.com/bblfsh/csharp-driver/issues/25

The issue was caused by a combination of factors:

1) `bblfshd` was limiting the timeout for driver startup to 5 sec. So the first request will always timeout if the driver init takes longer, and there is no way for the client to change it. It was fixed by passing the context directly.

2) If the first request failed, `bblfshd` was setting the driver pool anyway, and it become broken since no one returns the failed driver back to the queue. This caused a deadlock for any consequent requests. Fixed by not setting the pool/queue on error.

3) Maximal gRPC backoff delay was set too high, leading to unnecessary delay when driver takes more time to init. Fixed by capping the delay to 1 sec (it will retry).

This should fix the https://github.com/bblfsh/csharp-driver/issues/25, although the container startup may still take up to 5-6 sec. This is a startup time of the container itself, even before the driver gets a chance to execute. Need to check what may be the reason for this.

Few more issue were discovered along the way:

4) Failed driver was not shutdown properly. May have caused other issues in `bblfshd`. Fixed by stopping the container on the error path.

5) `bblfshd` was not acquiring the pool lock then shutting down. This may have caused data races with clients sending parse requests. Fixed by acquiring the lock.

Also, this PR includes few more changes:

1) Mutex for the driver pools was changed to RW to allow multiple callers to use an driver existing pool if it was already initialized.

2) `bblfshd` now supports a `--pprof` flag to expose a pprof HTTP endpoint. Useful to find deadlock, profile, etc.

### DIFF
--- a/cmd/bblfshd/main.go
+++ b/cmd/bblfshd/main.go
@@ -5,11 +5,14 @@ import (
 	"flag"
 	"fmt"
 	"net"
+	"net/http"
 	"os"
 	"os/signal"
 	"sync"
 	"syscall"
 	"time"
+
+	_ "net/http/pprof"
 
 	"github.com/bblfsh/bblfshd/daemon"
 	"github.com/bblfsh/bblfshd/runtime"
@@ -46,6 +49,10 @@ var (
 		format *string
 		fields *string
 	}
+	pprof struct {
+		enabled *bool
+		address *string
+	}
 	cmd *flag.FlagSet
 
 	usrListener net.Listener
@@ -66,6 +73,8 @@ func init() {
 	log.level = cmd.String("log-level", "info", "log level: panic, fatal, error, warning, info, debug.")
 	log.format = cmd.String("log-format", "text", "format of the logs: text or json.")
 	log.fields = cmd.String("log-fields", "", "extra fields to add to every log line in json format.")
+	pprof.enabled = cmd.Bool("profiler", false, "run profiler http endpoint (pprof).")
+	pprof.address = cmd.String("profiler-address", ":6060", "profiler address to listen on.")
 	cmd.Parse(os.Args[1:])
 
 	buildLogger()
@@ -100,6 +109,15 @@ func installRecommended(d *daemon.Daemon) error {
 
 func main() {
 	logrus.Infof("bblfshd version: %s (build: %s)", version, build)
+
+	if *pprof.enabled {
+		logrus.Infof("running pprof on %s", *pprof.address)
+		go func() {
+			if err := http.ListenAndServe(*pprof.address, nil); err != nil {
+				logrus.Errorf("cannot start pprof: %v", err)
+			}
+		}()
+	}
 
 	c, err := jaegercfg.FromEnv()
 	if err != nil {
--- a/daemon/daemon.go
+++ b/daemon/daemon.go
@@ -179,7 +179,7 @@ func (d *Daemon) newDriverPool(rctx context.Context, language string, image runt
 			return nil, err
 		}
 
-		if err := driver.Start(); err != nil {
+		if err := driver.Start(rctx); err != nil {
 			return nil, err
 		}
 
--- a/daemon/daemon.go
+++ b/daemon/daemon.go
@@ -29,7 +29,7 @@ type Daemon struct {
 	runtime   *runtime.Runtime
 	driverEnv []string
 
-	mu   sync.Mutex
+	mu   sync.RWMutex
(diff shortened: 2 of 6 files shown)


cmd/bblfshd/main.go
```diff
@@ -66,6 +69,7 @@ func init() {
 	log.level = cmd.String("log-level", "info", "log level: panic, fatal, error, warning, info, debug.")
 	log.format = cmd.String("log-format", "text", "format of the logs: text or json.")
 	log.fields = cmd.String("log-fields", "", "extra fields to add to every log line in json format.")
+	runPProf = cmd.Bool("pprof", false, "run pprof http endpoint.")
```
bzz: I remember already adding `pprof` support for bblfshd and indeed https://github.com/bblfsh/bblfshd/commit/b1b2883dbdc712570d16ce3edf1345162a6acf2f - it seems to be gone for some reason.

@dennwc do you think we could restore descriptive CLI flags in here please? Exposing `profiler` and `profiler-address`?

bzz:
CI seems to be 🔴 for some reason.

daemon/driver.go
```diff
@@ -132,19 +132,24 @@ func (i *DriverInstance) Start() error {
...
 		grpc.WithDialer(func(addr string, t time.Duration) (net.Conn, error) {
 			return net.DialTimeout("unix", addr, t)
 		}),
+		// always wait for the connection to become active
 		grpc.WithBlock(),
-		grpc.WithTimeout(ConnectionTimeout),
+		// we want to know sooner rather than later
+		// TODO(dennwc): sometimes the initialization of the container takes >5 sec
+		//               meaning that the time between Container.Start and the actual
+		//               execution of a Go server (not the native driver) takes this long
```
bzz: Just curious, could you please help me understand where would this time be visible in our current instrumentation?

cmd/bblfshd/main.go
dennwc: Done!

daemon/driver.go
dennwc: It will be included in the `bblfshd.pool.newDriverPool` span in Jaeger.

But there is no direct measurement for this specific delay - it will account for the whole "start driver, wait for the gRPC server" step. I plan to add more instrumentation later because it requires changes to SDK and drivers.

dennwc:
CI is green

cmd/bblfshd/main.go
```diff
@@ -5,6 +5,8 @@ import (
 	"flag"
 	"fmt"
 	"net"
+	"net/http"
+	_ "net/http/pprof"
```
creachadair: I suggest grouping the blank (side-effecting) imports separately, below the others.

cmd/bblfshd/main.go
```diff
@@ -101,6 +109,16 @@ func installRecommended(d *daemon.Daemon) error {
 func main() {
 	logrus.Infof("bblfshd version: %s (build: %s)", version, build)
 
+	if *pprof.enabled {
+		addr := *pprof.address
```
creachadair: Do we need the extra variable?

daemon/daemon.go
```diff
@@ -131,10 +131,17 @@ func (d *Daemon) RemoveDriver(language string) error {
...
+	d.mu.RUnlock()
+	if ok {
+		return dp, nil
+	}
+
 	d.mu.Lock()
 	defer d.mu.Unlock()
-
-	if dp, ok := d.pool[language]; ok {
+	dp, ok = d.pool[language]
```
creachadair: This is a little subtle, and probably deserves a comment: If you reach this point the original version of `dp` you read at line 135 is actually fine—but it's possible it has since been evicted (between releasing the reader lock and acquiring the writer lock). It isn't clear why you couldn't just _use_ `dp` at that point—but I assume the reason is that you want to make sure you never use a driver that isn't mapped?

creachadair: Actually no, I take it back: `dp` is _not_ fine at this point.

dennwc: Correct. Should I write the comment for it? I assumed it's more a less a common pattern.

creachadair: No need, it's fine. It's not that common—but it's true that upgrading a lock is certainly a reasonable thing to do and can't be done in one step.

PR #251, opened 2019-02-13
Always map enry->bblfsh language names

bzz:
This fixes https://github.com/bblfsh/client-go/issues/108

Before, we were mapping enry->bblfsh language name only in case it was detected by bblfshd itself (by calling to enry).

Now same change is applied on any user input, as the main use case for the clients is to detect language with Enry first and then call bblfshd with it, but language names are different (e.g `C#` vs `csharp`)

This is a workaround, until proper multiple `language_aliaces` are supported by every driver in it's manifest.

TEST PLAN:
 -  `$ bblfsh-cli -l 'C#'./some.cs`

### DIFF
--- a/daemon/language.go
+++ b/daemon/language.go
@@ -18,7 +18,7 @@ func GetLanguage(filename string, content []byte) string {
 }
 
 // normalize maps enry language names to the bblfsh ones.
-// TODO(bzz): remove this as soon as languafe_aliaces are supported in bblfsh
+// TODO(bzz): remove this as soon as language aliases are supported in bblfsh
 // driver manifest.
 func normalize(languageName string) string {
 	lang := strings.ToLower(languageName)
--- a/daemon/language.go
+++ b/daemon/language.go
@@ -14,7 +14,14 @@ func GetLanguage(filename string, content []byte) string {
 		return lang
 	}
 
-	lang = strings.ToLower(lang)
+	return normalize(lang)
+}
+
+// normalize maps enry language names to the bblfsh ones.
+// TODO(bzz): remove this as soon as languafe_aliaces are supported in bblfsh
+// driver manifest.
+func normalize(languageName string) string {
+	lang := strings.ToLower(languageName)
 	lang = strings.Replace(lang, " ", "-", -1)
 	lang = strings.Replace(lang, "+", "p", -1)
 	lang = strings.Replace(lang, "#", "sharp", -1)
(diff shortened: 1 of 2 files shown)


daemon/language.go
```diff
@@ -14,7 +14,14 @@ func GetLanguage(filename string, content []byte) string {
 		return lang
 	}
 
-	lang = strings.ToLower(lang)
+	return normalize(lang)
+}
+
+// normalize maps enry language names to the bblfsh ones.
+// TODO(bzz): remove this as soon as languafe_aliaces are supported in bblfsh
```
creachadair:
```suggestion
// TODO(bzz): remove this as soon as language aliases are supported in bblfsh
```

daemon/service.go
```diff
@@ -102,6 +102,8 @@ func (s *ServiceV2) selectPool(rctx context.Context, language, content, filename
 			return "", nil, err
 		}
 		language = lang
```
creachadair: Does this one need a `normalize` call too?

bzz: This one will already have it normalized after `s.detectLanguage` that calls it underneath

daemon/language.go
```diff
@@ -14,7 +14,14 @@ func GetLanguage(filename string, content []byte) string {
...
+}
+
+// normalize maps enry language names to the bblfsh ones.
+// TODO(bzz): remove this as soon as language aliases are supported in bblfsh
+// driver manifest.
+func normalize(languageName string) string {
+	lang := strings.ToLower(languageName)
 	lang = strings.Replace(lang, " ", "-", -1)
 	lang = strings.Replace(lang, "+", "p", -1)
 	lang = strings.Replace(lang, "#", "sharp", -1)
```
dennwc: This is an interesting way of solving this :)

Do you think making it an explicit map makes sense?

bzz:
> it may be necessary to list those aliases in `SupportedLanguages` as well. What do you think?

I think this would be out of scope of this PR/Issue - it only supposed to allow of consumption of enry names, as requested.

And AFAIK this is not even necessary - SupportedLanguages already return bblfsh names.

As [discussed elsewhere](https://src-d.slack.com/archives/C7UDG9VNY/p1550071280093200) the consensus is to add this workaround until we have a proper multiple aliases support.

daemon/language.go
bzz: > This is an interesting way of solving this :)

Well, it's the smallest possible change consistent with current behavior :)

As this is a workaround, I would prefer to keep this change as least invasive as possible. 

Given that we do not have much tests for it and that the current approach already works (TM) for API calls \wo the language -  I would postponed changing implementation until proper alias support.

dennwc:
OK, let's do it!

PR #254, opened 2019-02-22
Fix the context passed to the driver factory

### LINKED ISSUE #253: Simple DoS on bblfhsd with error re-scaling pool: context canceled
Parsing this syntactically incorrect js fragment triggers https://github.com/bblfsh/sdk/issues/371 but if a `bblfshd` is used instead of the driver directly - this condition kick the `bblfhsd v2.11.7` into the infinite loop on the next request to the same driver

Logs
```
time="2019-02-21T21:11:10Z" level=error msg="error re-scaling pool: context canceled" language=javascript
time="2019-02-21T21:11:10Z" level=debug msg="scaling driver pool from 0 instance(s) to 1 instance(s)" language=javascript
time="2019-02-21T21:11:10Z" level=debug msg="spawning driver instance "bblfsh/javascript-driver:latest" ..."
time="2019-02-21T21:11:10Z" level=error msg="error re-scaling pool: context canceled" language=javascript
time="2019-02-21T21:11:10Z" level=debug msg="scaling driver pool from 0 instance(s) to 1 instance(s)" language=javascript
time="2019-02-21T21:11:10Z" level=debug msg="spawning driver instance "bblfsh/javascript-driver:latest" ..."
```

Could be related to #252 

## Steps to reproduce

Same as in https://github.com/bblfsh/sdk/issues/371 but with `bblfshd` instead of only the driver.

dennwc:
Previous changes added a context argument to newDriverPool, which is called when the first request for the driver is received. By mistake, the context of this first request is captured by the closure and is used to create all other driver instances. Thus, if the first request specifies the context timeout and fails for some reason, all future requests will fail with "context canceled" error.

Instead, we need to pass the real context from the user request down to the pool's driver factory function.

This is not an ideal way of managing drivers - pool should be managed by a separate goroutine. This will be implemented later. This is merely a hotfix for the real issue.

Fixes #253

### DIFF
--- a/daemon/common_test.go
+++ b/daemon/common_test.go
@@ -26,7 +26,7 @@ type mockDriver struct {
 	MockStatus  protocol.Status
 }
 
-func newMockDriver() (Driver, error) {
+func newMockDriver(ctx context.Context) (Driver, error) {
 	return &mockDriver{
 		MockID:     runtime.NewULID().String(),
 		MockStatus: protocol.Running,
@@ -63,7 +63,7 @@ func (d *mockDriver) Stop() error {
 }
 
 func newEchoDriver() *echoDriver {
-	d, _ := newMockDriver()
+	d, _ := newMockDriver(context.Background())
 	return &echoDriver{
 		Driver: d,
 	}
--- a/daemon/daemon.go
+++ b/daemon/daemon.go
@@ -173,11 +173,11 @@ func (d *Daemon) getDriverImage(rctx context.Context, language string) (runtime.
 
 // newDriverPool, instance a new driver pool for the given language and image
 // and should be called under a lock.
-func (d *Daemon) newDriverPool(rctx context.Context, language string, image runtime.DriverImage) (*DriverPool, error) {
-	sp, _ := opentracing.StartSpanFromContext(rctx, "bblfshd.pool.newDriverPool")
+func (d *Daemon) newDriverPool(ctx context.Context, language string, image runtime.DriverImage) (*DriverPool, error) {
+	sp, _ := opentracing.StartSpanFromContext(ctx, "bblfshd.pool.newDriverPool")
 	defer sp.Finish()
 
-	dp := NewDriverPool(func() (Driver, error) {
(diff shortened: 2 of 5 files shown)


creachadair:
Can you add a bit more…uh…context, to the commit description?

I ask because this change looks structurally reasonable, but I think we need to be careful _not_ to use a request-specific context for pool maintenance. This is a long-running task that is not request-specific, and is one of the unusual cases where it might make sense for the pool to capture and hold a Context value for a long time. Most of the time that will be the background context, but we do want the main package to be able to control shutdown, so I wouldn't hard-wire it to that.

That said: I'm not sure that in this case we actually want to plumb the context through the request flow—it might actually be simpler (or perhaps not) to store it at construction.

dennwc:
@creachadair I agree that passing the context directly is not an ideal solution. We definitely need to rewrite the pool to manage instances on a separate goroutine, separately from user requests.

But it will require more work and time, while I think this issue blocks one of the PoCs to a certain degree.

The change should work reasonably well - it will create driver instance on the user request, as it was doing before, but it will allow canceling the creation if the user sets a short deadline. But yes, we will waste time to start and kill the driver. Looking at the code, it requires significantly more effort to rewrite it to be separate. I'm working on it right now.

creachadair:
> @creachadair I agree that passing the context directly is not an ideal solution. We definitely need to rewrite the pool to manage instances on a separate goroutine, separately from user requests.
> 
> But it will require more work and time, while I think this issue blocks one of the PoCs to a certain degree.

I agree, and I don't think we should block this PR for that. But if you're willing to add some more of the details—of why we're doing this, now, and what will need to be done to make it better—in the PR description and possibly as comments, that would help.
 
> The change should work reasonably well - it will create driver instance on the user request, as it was doing before, but it will allow canceling the creation if the user sets a short deadline. But yes, we will waste time to start and kill the driver. Looking at the code, it requires significantly more effort to rewrite it to be separate. I'm working on it right now.

Yeah, it is definitely a bigger project, and I don't see any problems with this solution for now. I'm mainly just asking for more docs. :)

dennwc:
Updated the commit message and the description. Ready for another round :)

PR #256, opened 2019-02-28
Validate that source files are encoded in UTF-8

### LINKED ISSUE #238: Parse RPC cannot handle file contents that are invalid UTF-8.
## Summary

This issue was reported by Fábio Costa Silva in the [source{d} community Slack](https://sourced-community.slack.com). What follows is my distillation of it and a relatively-minimal repro.

The [`ParseRequest`](https://github.com/bblfsh/sdk/blob/42e2023d304c66a242444d467c48f6913748e411/protocol/driver.proto#L14) protobuf message specifies a `content` field of type `string`. Per the [protobuf language guide](https://developers.google.com/protocol-buffers/docs/proto3#scalar),

> A string must always contain UTF-8 encoded or 7-bit ASCII text.

If a file is encoded in some other way, e.g., ISO-8859-1 (Latin 1), the RPC layer will generate an error marshaling the request message and fail.

## Repro

```shell
# One-time setup
mkdir repro
cd repro
git init
git commit -m "Initial commit." --allow-empty
python -c "print u'package bad;\npublic class Bad {\n  static final String bad = \"\\xe9\";\n}\n'.encode('latin-1')" > Bad.java
git add Bad.java
git commit -m "Add a non-UTF-8 encoded Java file." .
srcd init

# Reproduction
srcd sql 'select uast(blob_content, language(file_path, blob_content)) from files;'
```
## Expected

A UAST for this syntactically-valid Java file.

## Observed

```
rpc error: code = Internal desc = grpc: error while marshaling: proto: invalid UTF-8 string
```

The correct fix is to convert this field to type `bytes`. That is technically a breaking change, but in this case it is a fairly innocuous one: Any text that would have been accepted by a `string` field will also be accepted by a `bytes` field, but not vice versa. So it is wire-safe, but it changes the API surface of the generated code.

dennwc:
As described in #238, all the driver pipeline expects file content to be UTF-8, however, we do not strictly enforce it.

This change adds additional checks in `Parse` methods to ensure that content is encoded correctly.

Fixes #238

### DIFF
--- a/daemon/errors.go
+++ b/daemon/errors.go
@@ -21,4 +21,6 @@ var (
 	ErrUnauthorized = errors.NewKind("unauthorized: authentication required to access %s (image: %s)")
 	// ErrLanguageDetection indicates that language was not detected by Enry.
 	ErrLanguageDetection = errors.NewKind("could not autodetect language")
+	// ErrUnknownEncoding is returned for parse requests with a file content in a non-UTF8 encoding.
+	ErrUnknownEncoding = errors.NewKind("unknown source file encoding (expected UTF-8)")
 )
--- a/daemon/service.go
+++ b/daemon/service.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"fmt"
 	"time"
+	"unicode/utf8"
 
 	"github.com/opentracing/opentracing-go"
 	"github.com/sirupsen/logrus"
@@ -39,6 +40,12 @@ func (s *ServiceV2) Parse(rctx xcontext.Context, req *protocol2.ParseRequest) (r
 	if req.Content == "" {
 		logrus.Debugf("empty request received, returning empty UAST")
 		return resp, nil
+	}
+
+	if !utf8.ValidString(req.Content) {
+		err := ErrUnknownEncoding.New()
+		logrus.Debugf("parse v2 (%s): %s", req.Filename, err)
+		return nil, err
 	}
 
 	language, dp, err := s.selectPool(ctx, req.Language, req.Content, req.Filename)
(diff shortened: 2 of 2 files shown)


daemon/errors.go
```diff
@@ -21,4 +21,6 @@ var (
 	ErrUnauthorized = errors.NewKind("unauthorized: authentication required to access %s (image: %s)")
 	// ErrLanguageDetection indicates that language was not detected by Enry.
 	ErrLanguageDetection = errors.NewKind("could not autodetect language")
+	// ErrInvalidEncoding is returned for parse requests that provide a file in non-UTF8 encoding.
```
bzz:
```suggestion
	// ErrInvalidEncoding is returned for parse requests with a file content in a non-UTF8 encoding.
```

daemon/service.go
```diff
@@ -41,6 +42,12 @@ func (s *ServiceV2) Parse(rctx xcontext.Context, req *protocol2.ParseRequest) (r
 		return resp, nil
 	}
 
+	if !utf8.ValidString(req.Content) {
+		err := ErrInvalidEncoding.New()
+		logrus.Debugf("parse v2: %s", err)
```
bzz: To assist the debugging here and below, do you think it could be beneficial to include a bit of the "context" of this request inside the error? E.g the file name or something similar that would later help to identify the failure source

daemon/errors.go
creachadair: I'm not sure we want to say that other encodings are "invalid", since in this case the file is not broken, it's merely not using the encoding we want. I worry the user might think we're saying the problem is in their data rather than a limitation of our system.

I suggest being more specific about the complaint here, e.g., `ErrUnknownEncoding = errors.NewKind("unknown source file encoding (expected UTF-8)")` or words to that effect. (After all, if we knew what it was, we could have transcoded it)

PR #257, opened 2019-02-28
Small refactorings

dennwc:
Some small refactorings to prepare for the pool changes. The PR includes no functional changes, only a few simplifications or more strict checks.

See separate commit description for the reasoning behind each change.

### DIFF
--- a/daemon/pool.go
+++ b/daemon/pool.go
@@ -370,9 +370,8 @@ func DefaultScalingPolicy() ScalingPolicy {
 
 type movingAverage struct {
 	ScalingPolicy
-	loads  []float64
-	pos    int
-	filled bool
+	loads []float64
+	pos   int
 }
 
 // MovingAverage computes a moving average of the load and forwards it to the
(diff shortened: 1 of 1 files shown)


daemon/pool.go
```diff
@@ -381,31 +380,28 @@ type movingAverage struct {
...
+		loads:         make([]float64, 0, window),
 		pos:           0,
-		filled:        false,
 	}
 }
 
 func (p *movingAverage) Scale(total, load int) int {
-	p.loads[p.pos] = float64(load)
+	if len(p.loads) < cap(p.loads) {
+		p.loads = append(p.loads, float64(load))
```
creachadair: One thing to be careful about here is that appending may (and generally will, when it has to) increase the capacity more than just the amount needed for the append:

```
$ cat <<EOF | goimports > t.go
package main

import "fmt"

func main() {
        t := make([]int, 0, 2)
        for i := 0; i < 10; i++ {
                t = append(t, i)
                fmt.Printf("append %d\tlen=%d\tcap=%d\n", i, len(t), cap(t))
        }
}
EOF
$ go run t.go
append 0        len=1   cap=2
append 1        len=2   cap=2
append 2        len=3   cap=4
append 3        len=4   cap=4
append 4        len=5   cap=8
append 5        len=6   cap=8
append 6        len=7   cap=8
append 7        len=8   cap=8
append 8        len=9   cap=16
append 9        len=10  cap=16
```

Maybe this is OK in this case, because you're being careful about when to append—but even if this path only happens in the "not yet full" case I'm not convinced we will wind up with a cap that respects the window.

daemon/pool.go
```diff
@@ -24,9 +24,8 @@ var (
 	// scaling policy (see DefaultScalingPolicy()).
 	DefaultMaxInstancesPerDriver = runtime.NumCPU()
 
-	ErrPoolClosed         = errors.NewKind("driver pool already closed")
-	ErrInvalidPoolTimeout = errors.NewKind("invalid timeout: %v")
-	ErrNegativeInstances  = errors.NewKind("cannot set instances to negative number")
+	ErrPoolClosed        = errors.NewKind("driver pool already closed")
```
creachadair: Please document these exported names!

daemon/pool.go
dennwc: https://github.com/bblfsh/bblfshd/pull/257/commits/4529634763beb4d7f32ce01228b20e440c52aa0d#diff-ef2ae0497174a079d805f1c50a0f2444R383

dennwc: It won't ever reallocate or overallocate if the capacity is enough.

creachadair: Ah, I see, this is _only_ done during the fill phase, so you won't ever append once the wrap marker moves. Ok, that makes sense.

daemon/pool.go
dennwc: Will do in the next PR.

bzz: It would make sense to include at least a doc for the new error in here

bzz: Did not realize https://github.com/bblfsh/bblfshd/pull/258/files#diff-ef2ae0497174a079d805f1c50a0f2444 is already a thing; 👍

PR #258, opened 2019-02-28
Rewrite driver pool

dennwc:
Based on #257. Only the last commit is relevant.

An old pool had some problems like leaking goroutines, leaving dead drivers to hang for too long and unpredictable scaling behavior.

The new driver pool introduced in this change should be safer and more predictable. It may also fix issues with the scaling that we discovered.

New implementation still runs a scaling goroutine that sets a target number of instances. But instead of directly applying this number by killing or starting instances, it delegates instance management to a separate goroutine.

This second goroutine also accepts requests for idle drivers. It listens on multiple event channels and it should be now clean from the code what decision it would make on each event.

Also, an old implementation was using a client's request context when starting new instances. In this implementation, a third goroutine is responsible for starting instances in the background.

### DIFF
--- a/daemon/pool.go
+++ b/daemon/pool.go
@@ -136,9 +136,19 @@ func (dp *DriverPool) Start(ctx context.Context) error {
 	dp.targetSize.Set(1)
 
 	dp.wg.Add(3)
-	go dp.runSpawn()
-	go dp.runPolicy()
-	go dp.manageDrivers()
+	go func() {
+		defer dp.wg.Done()
+		dp.runSpawn()
+	}()
+	go func() {
+		defer dp.wg.Done()
+		dp.runPolicy()
+	}()
+	go func() {
+		defer close(dp.stopped)
+		defer dp.wg.Done()
+		dp.manageDrivers()
+	}()
 
 	// wait for a single instance to come up
 	d, err := dp.getDriver(ctx)
@@ -155,7 +165,6 @@ func (dp *DriverPool) Start(ctx context.Context) error {
 // runPolicy goroutine re-evaluates the scaling policy on a regular time interval and sets
 // a target number of instances. The scaling itself will be performed by the manager goroutine.
 func (dp *DriverPool) runPolicy() {
-	defer dp.wg.Done()
 	ticker := time.NewTicker(time.Millisecond * 100)
 	defer ticker.Stop()
 	for {
@@ -220,7 +229,6 @@ func (dp *DriverPool) spawnOne() {
 
 // runSpawn is a goroutine responsible for spawning new instances in the background.
 func (dp *DriverPool) runSpawn() {
-	defer dp.wg.Done()
 	for {
 		select {
 		case <-dp.stop:
@@ -455,8 +463,6 @@ func (dp *DriverPool) drain() {
 // It will accept all client requests for drivers if there are no drivers in the idle state.
 // It will also take care of draining instances when the pool closes.
 func (dp *DriverPool) manageDrivers() {
-	defer close(dp.stopped)
-	defer dp.wg.Done()
 	defer dp.drain()
 
 	for {
--- a/daemon/pool.go
+++ b/daemon/pool.go
@@ -655,6 +655,11 @@ func (dp *DriverPool) getDriver(rctx context.Context) (Driver, error) {
 	sp, ctx := opentracing.StartSpanFromContext(rctx, "bblfshd.pool.getDriver")
 	defer sp.Finish()
 
+	if dp.poolCtx == nil {
+		// not running
+		return nil, ErrPoolClosed.New()
+	}
+
 	for {
 		d, err := dp.getIdle(ctx)
 		if ErrPoolClosed.Is(err) {
--- a/daemon/pool.go
+++ b/daemon/pool.go
@@ -190,7 +190,9 @@ func (dp *DriverPool) runPolicy(ctx context.Context) {
 
 		target := dp.ScalingPolicy.Scale(total, load-idle)
 		if target < 1 {
-			target = 1 // there should be always at least 1 instance
+			// there should be always at least 1 instance
+			// TODO(dennwc): policies must never return 0 instances
+			target = 1
 		}
 		old := dp.targetSize.Set(target)
 		if old != target {
@@ -319,10 +321,29 @@ func (dp *DriverPool) scale() {
 	if dn < 0 {
 		// scale down
 		for i := 0; i < -dn; i++ {
+			select {
+			case <-stop:
+				return
+			case req := <-dp.get:
+				// no idle drivers, and there is a client waiting for us
+				// do the scaling "inline" while serving the request
+				dp.rescaleLater(req)
+				return
+			case d := <-dp.put:
+				// prefer to kill driver that are returned by clients instead an idle ones
+				// idle map may be accessed without management goroutine, thus it's more
+				// valuable to keep it full
+				dp.killDriver(d)
+				continue
+			default:
+			}
+			// only idle drivers remain - start killing those
 			if d, ok := dp.peekIdle(); ok {
 				dp.killDriver(d)
 				continue
 			}
+			// no drivers are idle, only way to downscale is to wait for clients to put
+			// their drivers back to the pool
 			select {
 			case <-stop:
 				return
--- a/daemon/pool.go
+++ b/daemon/pool.go
@@ -64,6 +64,7 @@ type DriverPool struct {
 
 	// rescale accepts signals passed from the runPolicy goroutine to the manager goroutine.
 	// It allows to re-evaluate scaling conditions when waiting for an idle driver.
+	// The channel must have a buffer and sends to this channel must be used with default.
 	rescale chan struct{}
 
 	// spawn accepts signals to the spawner goroutine to run a new driver.
@@ -131,10 +132,15 @@ func (dp *DriverPool) Start(ctx context.Context) error {
 	// But an alternative is to re-implement a root Context, which is even worse.
 	dp.poolCtx, dp.stop = context.WithCancel(context.Background())
 
+	// This channel is read by the pool manager goroutine as a signal to rescale.
+	// The scaling policy goroutine write to this channel whenever the allow number of
+	// drivers changes. The channel must have a buffer of at least 1 and sends to the
+	// channel should not block.
+	dp.rescale = make(chan struct{}, 1)
+
 	dp.stopped = make(chan struct{})
 	dp.spawn = make(chan struct{})
 	dp.spawnErr = make(chan error)
-	dp.rescale = make(chan struct{}, 1)
 	dp.get = make(chan driverRequest)
 	dp.put = make(chan Driver)
 	dp.drivers.idle = make(map[Driver]struct{})
@@ -196,8 +202,12 @@ func (dp *DriverPool) runPolicy(ctx context.Context) {
 		}
 		old := dp.targetSize.Set(target)
 		if old != target {
-			// optionally signal to the manager goroutine
+			// send a signal to the manager goroutine
 			select {
+			// the channel has buffer of 1 so it acts like a deferred signal
+			// the send will fail only if the channel is already full, meaning
+			// that manager goroutine haven't had time to receive the previous
+			// signal yet, and it's ignore the signal we are trying to send
 			case dp.rescale <- struct{}{}:
 			default:
 			}
--- a/daemon/pool.go
+++ b/daemon/pool.go
@@ -47,8 +47,12 @@ type DriverPool struct {
 	// wg tracks all goroutines owned by the driver pool.
 	wg sync.WaitGroup
 
-	// stop channel should be closed to send a stop signal to the pool.
-	stop chan struct{}
+	// poolCtx will be cancelled as a signal that the pool is closing.
+	poolCtx context.Context
+	// stop is called to send a stop channel to the pool. stopped channel will be closed
+	// when the pool is fully stopped.
+	stop func()
+
 	// stopped is closed when the pool (and the manager goroutine) stops.
 	stopped chan struct{}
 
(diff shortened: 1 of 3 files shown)


daemon/pool.go
```diff
@@ -402,26 +402,26 @@ func (p *movingAverage) Scale(total, load int) int {
 	}
 
 	avg := sum / float64(len(p.loads))
-	return p.ScalingPolicy.Scale(total, int(avg))
+	return p.sub.Scale(total, int(avg))
 }
 
 type minMax struct {
-	ScalingPolicy
+	Sub      ScalingPolicy
```
creachadair: I'm not entirely clear on why the name of this field is `sub`. When it was unexported it didn't really matter, but now that it's exported it probably deserves either a comment or maybe a more descriptive name. :)

dennwc: Unexported and commented on the field.

daemon/pool.go
```diff
@@ -73,123 +100,445 @@ func NewDriverPool(factory FactoryFunction) *DriverPool {
...
+// scale the driver pool to the current target number of instances.
+func (dp *DriverPool) scale() {
+	dn := dp.scaleDiff()
+	if dn == 0 {
+		return
+	}
+
+	// rescaleLater interrupts the scaling and serves the request first.
+	// It will make sure to continue scaling later.
+	rescaleLater := func(req driverRequest) {
```
creachadair: Can this be a method of `*DriverPool`? It doesn't seem to close over anything else in the environment.

daemon/pool.go
```diff
@@ -73,123 +100,445 @@ func NewDriverPool(factory FactoryFunction) *DriverPool {
...
+				dp.killDriver(d)
+				continue
+			}
+			select {
+			case <-dp.stop:
+				return
+			case req := <-dp.get:
+				rescaleLater(req)
+				return
+			case d := <-dp.put:
```
creachadair: This is clever enough that it probably deserves a comment: If someone is trying to put back a driver while we're scaling down, we'll kill theirs rather than sacrificing an idle one. 

That seems good, but I wonder whether we shouldn't do this check _before_ we check for idles?

daemon/pool.go
```diff
@@ -73,123 +100,445 @@ func NewDriverPool(factory FactoryFunction) *DriverPool {
...
+	// rescaleLater interrupts the scaling and serves the request first.
+	// It will make sure to continue scaling later.
+	rescaleLater := func(req driverRequest) {
+		select {
+		case dp.rescale <- struct{}{}:
+		default:
+		}
+		dp.waitOrScale(req)
+	}
+	if dn < 0 {
```
creachadair: Maybe consider splitting this into `dp.scaleUp` and `dp.scaleDown`?

daemon/pool.go
```diff
@@ -73,123 +100,445 @@ func NewDriverPool(factory FactoryFunction) *DriverPool {
...
+			i--
+			dn = dp.scaleDiff()
+		case dp.spawn <- struct{}{}:
+			// spawn request sent to the spawn goroutine
+			select {
+			case <-dp.stop:
+				return
+			case d := <-dp.put:
+				dp.setIdle(d)
+			case <-dp.spawnErr:
```
creachadair: If it's already been logged and we aren't going to do anything about it, can we not bother sending it at all?

daemon/pool.go
```diff
@@ -73,123 +100,445 @@ func NewDriverPool(factory FactoryFunction) *DriverPool {
...
+// wait for an instance to become available, or scale the pool down. The function will
+// block until the request is served or cancelled.
+func (dp *DriverPool) waitOrScale(req driverRequest) {
+	// Note that we don't care about the pool closing in this function.
+	// If it happens, the client will receive this signal before we do and will
+	// cancel the request anyway. This reduces the number of select statements.
+
+	// returnDriver will either return the driver to the client,
+	// or will return it to the idle driver queue.
+	returnDriver := func(d Driver) {
```
creachadair: Can this be a method of `*DriverPool`?

```go
func (dp *DriverPool) returnDriver(req driverRequest) { … }
```

It does close over `req` but it doesn't really need to, since it doesn't outlive this scope.

daemon/pool.go
```diff
@@ -73,123 +100,445 @@ func NewDriverPool(factory FactoryFunction) *DriverPool {
...
 
-	return nil
-}
+	// loop allows to re-evaluate scaling conditions
+	for {
+		dn := dp.scaleDiff()
+
+		if dp.running.Value()+dp.spawning.Value() == 0 {
+			if dn < 0 {
+				// This shouldn't really happen: pool is draining, but there are no running
```
creachadair: Do you mean "shouldn't" in the sense that it's not common, or in the sense that if it happens something is broken in the upstream logic? ("Uncommon" is fine; "broken" is more serious).

daemon/pool.go
```diff
@@ -73,123 +100,445 @@ func NewDriverPool(factory FactoryFunction) *DriverPool {
...
+	ctx, cancel := context.WithCancel(rctx)
+	defer cancel()
+
+	resp := make(chan Driver)
+	errc := make(chan error, 1)
+	req := driverRequest{
+		out: resp, err: errc, cancel: ctx.Done(),
+	}
+	select {
+	case <-req.cancel: // parent context cancelled
```
creachadair: Do you really mean the parent context of `rctx`? If so, I don't understand why we are checking `req.cancel` instead of `ctx.Done()` here: Isn't `req.cancel` a signal from the requestor rather than the context? 

I'm assuming I have missed something.

daemon/pool.go
```diff
@@ -475,3 +771,31 @@ func (p *aimd) Scale(total, load int) int {
 
 	return res
 }
+
+// chanContext wraps a receive-only channel to act as a context.
```
creachadair: Please let's try to find a solution that doesn't require us to re-implement `context.Context` here. If we really need a separate context, we should—if at all possible—use the one from the `context` package. Instead of creating our own channel.

Can we use:

```go
ctx, cancel := context.WithCancel(context.Background())
```

to get a fresh context, and cache those values for use by the manager?  That gives us to cancel and check for cancellation, and would make it easy to plumb in deadlines if we needed to later.

The Go team strongly discourages re-implementing this interface (it's only an interface because there are different concrete wrappers in its own package for handling the wrapper semantics).

dennwc: Haven't known that reimplementing it is discouraged. Will use `Background` instead.

daemon/pool.go
dennwc: It's exactly the same, but avoids a call to `Done()`. Added the comment.

daemon/pool.go
dennwc: This is more like a "broken" case - the scaling policy should not return values that lead to this state. I was more inclined to panic here, but it may be a bit too much (it will kill bblfshd instance).

daemon/pool.go
dennwc: It is unused only in this code path. The reason for this channel is that an error may or may not persist. If it does, we are not worried about receiving a specific one and may drop some of them when it's received by a background goroutine. But at the same time, we can return it to the client if he asks for a new driver. And if an error does not persist, it will only be logged and the client won't notice any disruption.

daemon/pool.go
dennwc: Yes, this makes total sense. Checking with `select` + `default` is not that reliable, but at least it will catch some cases. Will have to duplicate the `select` thought.

daemon/pool.go
```diff
@@ -73,123 +100,445 @@ func NewDriverPool(factory FactoryFunction) *DriverPool {
...
+		idle := len(dp.drivers.idle)
+		dp.drivers.RUnlock()
+
+		target := dp.ScalingPolicy.Scale(total, load-idle)
+		if target < 1 {
+			target = 1 // there should be always at least 1 instance
+		}
+		old := dp.target.Set(target)
+		if old != target {
+			// optionally signal to the manager goroutine
```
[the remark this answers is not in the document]
dennwc: The `rescale` channel is buffered, thus this will just set a "flag" for the manager goroutine to re-evaluate scaling conditions later.

The benefits that if the pool is starving, and the policy doesn't allow to scale up, the manager goroutine will be blocked waiting for drivers to be put back to the pool. But the scaling policy is not static, it may allow more instances once more requests are waiting for the driver. In this case, this send will unblock the manager goroutine.

The previous implementation I did a few days ago inlined the policy re-evaluation here. But it turned out to be too complex. The policy expects some stable "tick" for re-evaluation, so it's easier to handle it in a separate goroutine.

daemon/pool.go
```diff
@@ -73,123 +100,445 @@ func NewDriverPool(factory FactoryFunction) *DriverPool {
...
-		return nil
-	}
-	if dn > 0 {
-		return dp.add(ctx, dn)
+// spawnOne starts a new driver instance. It will keep trying to run it in case of a failure.
+func (dp *DriverPool) spawnOne() {
+	dp.spawning.Add(1)
+	defer dp.spawning.Add(-1)
+
+	// TODO(dennwc): use exponential backoff instead?
```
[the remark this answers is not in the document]
dennwc: I cannot think of any other good way to do it without a backoff.

If there is a fixed upper bound on retries, it may quickly spend all those retry attempts and kill bblfshd because of the panic.

In case we increase the delay between retries, it may be too slow to retry and requests will be waiting for this delay even on single failure.

So I guess I should add the backoff and the upper bound at the same time. Maybe in the next PR? Or should I do it here?

daemon/pool.go
```diff
@@ -73,123 +100,445 @@ func NewDriverPool(factory FactoryFunction) *DriverPool {
...
+		case <-ticker.C:
+		}
+		total := dp.running.Value()
+		load := dp.requests.Value()
+		dp.drivers.RLock()
+		idle := len(dp.drivers.idle)
+		dp.drivers.RUnlock()
+
+		target := dp.ScalingPolicy.Scale(total, load-idle)
+		if target < 1 {
```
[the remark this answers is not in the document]
dennwc: Added the comment to not forget it

dennwc:
Addressed all the feedback, ready for another round!

daemon/pool.go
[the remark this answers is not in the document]
creachadair: I agree it doesn't make sense to replicate policy evaluation here. My comment was meant more to determine whether we could just get rid of this extra "attempt" to flag the conditions: Since the manager goroutine already has a retry clock, and will retry anyway, it seems to me whether we flag it here or wait for the next tick can't materially change the response the manager has to a starved pool, right?

daemon/pool.go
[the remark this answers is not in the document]
creachadair: I think it's fine to defer it.

daemon/pool.go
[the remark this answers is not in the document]
bzz: sounds like reasonable simplification suggestion

dennwc:
@creachadair I think I know what you mean. Currently, the manager goroutine was no retry clock. It uses the `rescale` channel to get signals from the policy goroutine that has such a clock. But nothing prevents us from having a second clock with the same tick rate in manager goroutine, and use it re-evaluate scaling conditions instead of waiting on `rescale`. Is this what you meant?

dennwc:
@bzz Nothing new for a scaling policy. Unfortunately, we lost the ability to track driver spawning from client requests that require it. But this is by design - it will now happen in the background.

Regarding test coverage, it's quite good already - it found a few bugs in my code already.

creachadair:
> @creachadair I think I know what you mean. Currently, the manager goroutine was no retry clock. It uses the `rescale` channel to get signals from the policy goroutine that has such a clock. But nothing prevents us from having a second clock with the same tick rate in manager goroutine, and use it re-evaluate scaling conditions instead of waiting on `rescale`. Is this what you meant?

Yes. Well, my original question was to determine whether the manager _does_ have a retry mechanism already built in. But I see now that it doesn't. Is it correct, then, that the reason this channel write is safely optional is that (a) we _know_ the channel has a non-zero buffer, and therefore (b) if our write fails it can only be because the buffer is already full, and therefore (c) the manager will check anyway, even though our write failed?

If the above is correct, can I ask you to please document that in a comment? That is clever, but very subtle, and involves dependencies between the point of construction (to ensure the channel is buffered), the point of consumption (the manager is written to assume it will always be notified by some producer), and the point of production (here and other places where the signalling occurs). 

I worry about a complex design like that because it would be really easy to mess it up: A case in point, anytime I see a buffered channel, I see it as a red flag for concurrency bugs. As a reader I'm wondering, "is this the right buffer size? Did the author _know_ this was the right buffer size or did they just add values until the tests stopped failing?" In this case I don't think those questions apply, but someone coming back to this code later might well think they can remove the buffering.

daemon/pool.go
```diff
@@ -64,6 +64,7 @@ type DriverPool struct {
 
 	// rescale accepts signals passed from the runPolicy goroutine to the manager goroutine.
 	// It allows to re-evaluate scaling conditions when waiting for an idle driver.
+	// The channel must have a buffer and sends to this channel should be used with default.
```
creachadair: s/should/must/

daemon/pool.go
```diff
@@ -131,10 +132,14 @@ func (dp *DriverPool) Start(ctx context.Context) error {
 	// But an alternative is to re-implement a root Context, which is even worse.
 	dp.poolCtx, dp.stop = context.WithCancel(context.Background())
 
+	// Channel acts like a deferred signal, so it should have a buffer.
```
creachadair: s/should/must/

Please also consider mentioning _who_ reads and writes this channel (the following may be wrong): "This channel is read by the pool manager goroutine as a signal to rescale. Senders write to this channel whenever an event occurs that may affect pool size, such as a driver being returned. The channel must have a buffer of at least 1 and sends to the channel should not block."

(Is there actually more than one place where this write happens? If so we might think about putting a method on it to ensure we always do the nonblocking send)

PR #261, opened 2019-03-08
Use exponential backoff for driver failures

dennwc:
Currently, the pool will keep trying to run the driver if it fails and will do it each 250 ms.

Instead, use an exponential backoff that starts from 500 ms delay and increases to 1 min. The backoff will also stop retrying after 15 min of failures and will close the pool automatically.

### DIFF
--- a/daemon/pool.go
+++ b/daemon/pool.go
@@ -10,6 +10,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/cenkalti/backoff"
 	"github.com/opentracing/opentracing-go"
 	"github.com/sirupsen/logrus"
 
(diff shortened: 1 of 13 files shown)


daemon/pool.go
```diff
@@ -245,7 +245,12 @@ func (dp *DriverPool) spawnOne() {
 		case <-stop:
 			return // cancel
 		case dp.spawnErr <- err:
-		case <-ticker.C:
+		case _, ok := <-ticker.C:
+			if !ok {
+				dp.Logger.Errorf("driver keeps failing, closing the pool; error: %v", err)
+				go dp.Stop()
```
creachadair: Why is this in a separate goroutine? If we're closing down anyway I don't think we care about efficiency, and presumably we want to actually ensure this happens before returning?

dennwc: `Stop` will [wait](https://github.com/bblfsh/bblfshd/blob/master/daemon/pool.go#L725) for all goroutines to return, including the spawn goroutine. Because of this, it should be done asynchronously.

Will add a comment explaining this.

dennwc: Done.

Issue #194, opened 2018-10-01
Drivers are downscaled too fast


smola:
I'm using a bblfshd with Go drivers, under quite intensive load. Driver instances quickly scale up to the maximum (8) and then downscale to the minimum (2). This happens every couple of seconds. It seems that the default scaling policy is too aggressive for downscaling.

Changing to something like `MovingAverage(20, MinMax(1, DefaultMaxInstancesPerDriver, AIMD(1, 0.8)))` could help (moving average 10 -> 20, that is, from 1 second to 2 seconds, and decrease factor from 0.5 to 0.8). This is just a guess, I didn't test.

On top of providing some more conservative defaults on downscaling, we could parameterize those with environment variables, so that we can easily experiment with those.

smola:
Note that this looks like the reason why failed cleanups generate such large amount of stale files, see https://github.com/bblfsh/bblfshd/issues/168

PR #262, opened 2019-03-08
Refactor scaling policies

dennwc:
The issue described in #194 seems to be unrelated to scaling policies and should be fixed by a pool rewrite.

This PR tunes scaling parameters to prevent fast downscaling and exposes all scaling parameters as environment variables.

### DIFF
--- a/daemon/pool.go
+++ b/daemon/pool.go
@@ -27,6 +27,12 @@ var (
 	//
 	// Can be changed by setting BBLFSHD_MAX_DRIVER_INSTANCES.
 	DefaultMaxInstancesPerDriver = mustEnvInt("BBLFSHD_MAX_DRIVER_INSTANCES", runtime.NumCPU())
+	// DefaultMinInstancesPerDriver is the minimal number of instances of
+	// the same driver which will be launched following the default
+	// scaling policy (see DefaultScalingPolicy()).
+	//
+	// Can be changed by setting BBLFSHD_MIN_DRIVER_INSTANCES.
+	DefaultMinInstancesPerDriver = mustEnvInt("BBLFSHD_MIN_DRIVER_INSTANCES", 1)
 
 	// ErrPoolClosed is returned if the pool was already closed or is being closed.
 	ErrPoolClosed = errors.NewKind("driver pool already closed")
@@ -37,10 +43,20 @@ var (
 )
 
 var (
-	policyDefaultWindow    = mustEnvDur("BBLFSHD_POLICY_WINDOW", time.Second*3)
-	policyDefaultTick      = mustEnvDur("BBLFSHD_POLICY_TICK", time.Millisecond*100)
-	policyDefaultMin       = mustEnvInt("BBLFSHD_POLICY_MIN", 1)
-	policyDefaultScale     = mustEnvInt("BBLFSHD_POLICY_SCALE_INC", 1)
+	// policyDefaultWindow is a window for the average function used in the default scaling
+	// policy. The window will be divided by policyDefaultTick intervals to calculate the
+	// size of the window buffer, so this should ideally be a multiple of policyDefaultTick.
+	policyDefaultWindow = mustEnvDur("BBLFSHD_POLICY_WINDOW", time.Second*3)
+	// policyDefaultTick is a tick rate for the goroutine that re-evaluates the scaling
+	// policy for each driver pool.
+	policyDefaultTick = mustEnvDur("BBLFSHD_POLICY_TICK", time.Millisecond*100)
+	// policyDefaultScale is a default increment for an additive increase scaling.
+	//
+	// See AIMD for more details.
+	policyDefaultScale = mustEnvInt("BBLFSHD_POLICY_SCALE_INC", 1)
+	// policyDefaultDownscale is a default multiplier for a multiplicative decrease downscaling.
+	//
+	// See AIMD for more details.
 	policyDefaultDownscale = mustEnvFloat("BBLFSHD_POLICY_DOWNSCALE_MULT", 0.25)
 )
 
@@ -805,7 +821,7 @@ type ScalingPolicy interface {
 
 // defaultScalingPolicy is the same as DefaultScalingPolicy, but has no window.
 func defaultScalingPolicy() ScalingPolicy {
-	min := policyDefaultMin
+	min := DefaultMinInstancesPerDriver
 	if min <= 0 {
 		min = DefaultMaxInstancesPerDriver
 	}
--- a/daemon/pool.go
+++ b/daemon/pool.go
@@ -5,7 +5,9 @@ package daemon
 import (
 	"context"
 	"math"
+	"os"
 	"runtime"
+	"strconv"
 	"sync"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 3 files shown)


daemon/pool.go
```diff
@@ -32,6 +36,31 @@ var (
 	errDriverStopped = errors.NewKind("driver stopped")
 )
 
+var (
+	policyDefaultWindow    = envIntOr("BBLFSHD_POLICY_WINDOW", 30)
```
creachadair: What are the units on this? (And perhaps more importantly, how will the user know?)

daemon/pool.go
```diff
@@ -32,6 +36,31 @@ var (
...
+	policyDefaultMin       = envIntOr("BBLFSHD_POLICY_MIN", 1)
+	policyDefaultScale     = envIntOr("BBLFSHD_POLICY_SCALE", 1)
```
creachadair: Is this a multiplier or a cap?

daemon/pool.go
```diff
@@ -32,6 +36,31 @@ var (
...
+	policyDefaultDownscale = envFloatOr("BBLFSHD_POLICY_DOWNSCALE", 0.25)
+)
+
+func envIntOr(env string, def int) int {
+	s := os.Getenv(env)
+	v, err := strconv.Atoi(s)
+	if err != nil {
+		return def
```
creachadair: Since this is done at startup, it is probably OK to panic/fatal for issues here. Or at least, to log the error rather than silently using the default. 

I'd probably make this:

```go
func mustEnvInt(env string, def int) int {
   s := os.Getenv(env)
   if s == "" {
      return def
   }
   v, err := strconv.Atoi(s)
   if err != nil {
      log.Fatalf("Invalid value for %q: %q", env, s)
   }
   return v
}
```

(the "must" prefix is similar to the C++ convention "…OrDie" in some codebases)

daemon/pool.go
```diff
@@ -747,21 +776,33 @@ func (c *atomicInt) Value() int {
 // ScalingPolicy specifies whether instances should be started or stopped to
 // cope with load.
 type ScalingPolicy interface {
-	// Scale takes the number of total instances and the load. The load is
-	// the number of request waiting or, there is none, it is a negative
-	// value indicating how many instances are ready.
-	Scale(total, load int) int
+	// Scale takes the number of total instances, idle instances and the load.
```
creachadair: Total meaning the total number of instances currently already active?

daemon/pool.go
```diff
@@ -747,21 +776,33 @@ func (c *atomicInt) Value() int {
...
+	// Idle value indicates how many instances are ready. The load is the number
+	// of request waiting.
+	Scale(total, idle, load int) int
```
creachadair: Maybe also add that `Scale` may return a value ≤ 0 to use the system default?

daemon/pool.go
```diff
@@ -747,21 +776,33 @@ func (c *atomicInt) Value() int {
...
+}
+
+// defaultScalingPolicy is the same as DefaultScalingPolicy, but has no window.
+func defaultScalingPolicy() ScalingPolicy {
+	min := policyDefaultMin
+	if min < 0 {
```
creachadair: Do you want `<= 0` here?

daemon/pool.go
dennwc: Added a `_INC` suffix.

daemon/pool.go
dennwc: Changed units, it's now a time interval, instead of multiple of tick rate.

daemon/pool.go
```diff
@@ -747,21 +795,36 @@ func (c *atomicInt) Value() int {
...
-	Scale(total, load int) int
+	// Scale takes the total number of active instances, number idle instances and the
+	// number of requests waiting to get a driver instance. Idle may not be zero even if
+	// number of waiting requests is non-zero.
+	// Scale returns the new target number of instances to keep running. This number must
+	// not be less than 1.
+	Scale(total, idle, waiting int) int
+}
+
+// defaultScalingPolicy is the same as DefaultScalingPolicy, but has no window.
```
creachadair: Can you say a bit more in the comment about why we would want this (i.e., no-window)?

daemon/pool.go
creachadair: I'm not sure what `_INC` means in this context—is that an additive increment for the AIMD policy?

daemon/pool.go
```diff
@@ -32,6 +36,50 @@ var (
 	errDriverStopped = errors.NewKind("driver stopped")
 )
 
+var (
+	policyDefaultWindow    = mustEnvDur("BBLFSHD_POLICY_WINDOW", time.Second*3)
+	policyDefaultTick      = mustEnvDur("BBLFSHD_POLICY_TICK", time.Millisecond*100)
+	policyDefaultMin       = mustEnvInt("BBLFSHD_POLICY_MIN", 1)
+	policyDefaultScale     = mustEnvInt("BBLFSHD_POLICY_SCALE_INC", 1)
+	policyDefaultDownscale = mustEnvFloat("BBLFSHD_POLICY_DOWNSCALE_MULT", 0.25)
```
bzz: Although all these are all private, but as soon as they are exposed though evn vars - I'm sure the next maintainer, as well as any person doing Ops, will be very grateful for any documented tips on what each of those are for.

daemon/pool.go
```diff
@@ -22,7 +24,9 @@ var (
 	// DefaultMaxInstancesPerDriver is the maximum number of instances of
 	// the same driver which can be launched following the default
 	// scaling policy (see DefaultScalingPolicy()).
-	DefaultMaxInstancesPerDriver = runtime.NumCPU()
+	//
+	// Can be changed by setting BBLFSHD_MAX_DRIVER_INSTANCES.
+	DefaultMaxInstancesPerDriver = mustEnvInt("BBLFSHD_MAX_DRIVER_INSTANCES", runtime.NumCPU())
```
bzz: Shall this be documented somehow on CLI level?

Not sure it it's worth to put in README or directly to the documentation site - WDYT?

bzz:
> This PR tunes scaling parameters to prevent fast downscaling

Question - would it make sense/be possible to include some kind of "simulation" script on synthetic data, plotting the expected policy outcomes in some from e.g a 2D graph/SVG?

Model like that may have some initial "cost" to produce it for the first time, but also would communicate expectations very well and serve as a foundation for further informed improvements of the algorithm. WDYT?

daemon/pool.go
dennwc: Sorry, but is this really necessary? Finding usages for this function clearly shows that there are only two of them: one that uses it in the `DefaultScalingPolicy` with a window, and the second one in the tests. There are few tests for the window itself, but there were none for this combination.

creachadair: Is this one here only for the test? If so, can we simply move it into the `_test.go` file so that it doesn't have to pollute the main package? (It's fine for the tests to add things to the package to support the test, if that's all that is really going on here).

daemon/pool.go
dennwc: Added "Environment variables" section to the README. Also exposed `BBLFSHD_MIN_DRIVER_INSTANCES`.

dennwc:
@bzz The simulation would be nice indeed, but this is a topic for another PR. For now, there are enough tests that assert how each policy should behave.

daemon/pool.go
creachadair: The test is in the same package, so it should not be possible to update one without the other: The test wouldn't compile. All I'm really saying is that we shouldn't put code that is only used for the test into the part of the package we ship to dependencies, if possible.

creachadair: We chatted about this in Slack: The separation is to make it easier to test, but is not only for the test (I was confused). We can leave this alone for now and think about how to test it more completely as we refine it.

PR #274, opened 2019-03-26
Switch from Dep to Go modules

dennwc:
Switch from Dep to Go modules and build with Go 1.12.

Modules builds are repeatable, so the PR also drops `vendor` directory.

Also updates the SDK to the latest version.


dennwc:
@creachadair I don't expect workflow to change. The user either pulls the Docker image and doesn't care about building from source, or uses `make build` which still works with modules. The only difference is that Go 1.11+ is now required.

creachadair:
I'm not worried about it, but how could this change decrease coverage?

dennwc:
I have no idea to be honest. The [details](https://codecov.io/gh/bblfsh/bblfshd/compare/031513f8bb5de5752d1bd1a2883a26cb16c7d5f7...ef394cc6073a0226e76fb2524fa993e95a866c3f) seems to suggest that only whitespace change to 2 lines occured. Their algorythm may use the total byte size to calculate the score, at least it's my guess.

creachadair:
> I have no idea to be honest. The [details](https://codecov.io/gh/bblfsh/bblfshd/compare/031513f8bb5de5752d1bd1a2883a26cb16c7d5f7...ef394cc6073a0226e76fb2524fa993e95a866c3f) seems to suggest that only whitespace change to 2 lines occured. Their algorythm may use the total byte size to calculate the score, at least it's my guess.

That's certainly consistent with the size of the effect. Which I suppose only goes to reinforce my view that coverage is a sloppy metric and only marginally useful. 😀

.travis.yml
```diff
@@ -1,10 +1,13 @@
 language: go
 
 go:
-  - '1.10.x'
+  - '1.12.x'
```
bzz: should not we be testing on 2 latest go version?

dennwc: According to our [guide](https://github.com/src-d/guide/blob/master/engineering/conventions/go.md#supported-go-versions), applications support one latest stable Go version.

dennwc:
Holding this off a bit. Will first cut a patch release with the updated drivers, and then the minor release with Go modules.

PR #280, opened 2019-04-19
Require JAEGER_AGENT_HOST to enable tracing

dennwc:
Multiple users reported issues with running `bblfshd` without tracing. Sometimes the tracing setup with default parameters will fail on some systems, even if a user hasn't changed any environment variables to enable tracing.

PR requires `JAEGER_AGENT_HOST` to be set to enable tracing. It should fix the issue for those users.

### DIFF
--- a/cmd/bblfshd/main.go
+++ b/cmd/bblfshd/main.go
@@ -122,13 +122,11 @@ func main() {
 	if os.Getenv("JAEGER_AGENT_HOST") != "" {
 		c, err := jaegercfg.FromEnv()
 		if err != nil {
-			logrus.Errorf("error configuring tracer: %s", err)
-			os.Exit(1)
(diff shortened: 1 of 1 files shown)


cmd/bblfshd/main.go
```diff
@@ -119,17 +119,19 @@ func main() {
...
-		os.Exit(1)
-	}
-	closer, err := c.InitGlobalTracer("bblfshd")
-	if err != nil {
-		logrus.Errorf("error configuring tracer: %s", err)
-		os.Exit(1)
+	if os.Getenv("JAEGER_AGENT_HOST") != "" {
+		c, err := jaegercfg.FromEnv()
+		if err != nil {
+			logrus.Errorf("error configuring tracer: %s", err)
```
creachadair: Is this not equivalent to `logrus.Fatalf(…)`?

Issue #286, opened 2019-05-03
bblfshd crashed and does not start after parsing millions of files


vmarkovtsev:
I ran massive file parsing on 4 machines:

* `typos-1.infra.mining.prod.srcd.host`
* `typos-2.infra.mining.prod.srcd.host`
* `typos-3.infra.mining.prod.srcd.host`
* `typos-4.infra.mining.prod.srcd.host`

The pipeline was the same as in https://github.com/bblfsh/bblfshd/issues/270#issuecomment-486272814 How I started bblfshd:

```
docker run -d --name bblfshd --privileged -p 9432:9432 bblfsh/bblfshd
docker exec -it bblfshd bblfshctl driver install --all
```

All the versions correspond to #281 

Those instances ran for several days and then crashed. **`docker start bblfshd` starts the server and it immediately stops.** Each of the 4 machines has the same symptoms. The free space is fine: 75% of a 200GB volume is free.

Here are the logs:

* [typos-1](https://drive.google.com/open?id=1-w14XgsthOTKBmTPnUY2zg5IBljTVUCR)
* [typos-2](https://drive.google.com/open?id=1gaCTopQV7ptmhhc6QyWqaWgeSzxE9lhu)
* [typos-3](https://drive.google.com/open?id=1-W_Ajs0WI5lZowrHaP7qy_xt8tS12KPs)
* [typos-4](https://drive.google.com/open?id=1wFjJBUi_cVIBQk4zGLhplOFlrUQRIBt2)

Each file is ~4 gigs uncompressed.

I noticed that #270 persists. The access to my machines can be granted by @rporres upon request.

dennwc:
```
error creating control listener: listen unix /var/run/bblfshctl.sock: bind: address already in use
```
Try removing the `/var/run/bblfshctl.sock` file on the server.

dennwc:
It also shows a giant stack trace at the end. From a quick look, it may have been affected by https://github.com/bblfsh/bblfshd/issues/264.

dennwc:
bblfshd v2.13.0 is out and should fix the crash, memory consumption and hopefully the CPU consumption as well.

Leaving this issue open, since we still need to solve an issue with the socket left after bblfshd crash.

PR #288, opened 2019-05-03
Recreate the control socket after a crash

dennwc:
As reported in #286, `bblfshd` will not be able to restart after a crash because of the following error:
```
error creating control listener: listen unix /var/run/bblfshctl.sock: bind: address already in use
```

This usually happens on OOM kill, when `bblfshd` has no chance to clean up the socket on shutdown.

This change checks for this specific error, tries to determine if the socket is in use, and if not - removes it and tries to listen on it again.

### DIFF
--- a/cmd/bblfshd/main.go
+++ b/cmd/bblfshd/main.go
@@ -199,6 +199,13 @@ func listenUser(d *daemon.Daemon) {
 
 func listenControl(d *daemon.Daemon) {
 	var err error
+	if *ctl.network == "unix" {
+		// Remove returns an error if file does not exists
+		// if it returns nil, we know the file existed, so bblfshd might have crashed
+		if err := os.Remove(*ctl.address); err == nil {
+			logrus.Warningf("control socket %s (%s) already exists", *ctl.address, *ctl.network)
+		}
+	}
 	ctlListener, err = net.Listen(*ctl.network, *ctl.address)
 	if err != nil {
 		logrus.Fatalf("error creating control listener: %s", err)


cmd/bblfshd/main.go
```diff
@@ -197,9 +197,48 @@ func listenUser(d *daemon.Daemon) {
...
+	case *net.OpError:
+		return unwrap(err.Err)
+	case *os.SyscallError:
+		return unwrap(err.Err)
+	}
+	return err
+}
+
+func isErrno(err error, eno syscall.Errno) bool {
+	err = unwrap(err)
```
creachadair: Minor thing: Let's either use a fresh name here, or inline:

```go
if e, ok := unwrap(err).(syscall.Errno); ok {
   return e == eno
}
// or
base := unwrap(err)
if e, ok := base.(syscall.Errno); ok …
```

cmd/bblfshd/main.go
```diff
@@ -197,9 +197,48 @@ func listenUser(d *daemon.Daemon) {
...
+}
+
+func isConnRefused(err error) bool {
+	return isErrno(err, syscall.ECONNREFUSED)
+}
+
 func listenControl(d *daemon.Daemon) {
 	var err error
 	ctlListener, err = net.Listen(*ctl.network, *ctl.address)
+	if isAddrInUse(err) && *ctl.network == "unix" {
```
creachadair: Do we need to do all this? This socket is inside the container, not on the host, right? So in principle we should not have multiple bblfshd instances trying to use it. If that's correct, we can blindly `os.Remove` the socket path before we attempt to bind, and trust that we're either a restart (if it exists) or the first (if not).

dennwc: You are right, it's _usually_ in the container. But we also got multiple requests for running it without the container, that is possible right now. So I think it's better to be on a safe side and assume some might run two of those by accident.

cmd/bblfshd/main.go
dennwc: Done

cmd/bblfshd/main.go
creachadair: Even if it's running outside the container, though, we don't support running multiple instances on the same host (that's implicit in the fact that we use the same address).

So I don't think this should be necessary even outside the container. If we're starting up and we have a Unix-domain socket, I think we can just remove it before binding. (I think that's what Stevens recommends in Unix Network Programming, and it seems to be fairly common)

dennwc: Sounds reasonable. Will change it.

dennwc: Done

bzz: I belive this comment below is relevant to this thread https://github.com/bblfsh/bblfshd/commit/f629bc9f5ea1ef7890b57223d78281a962f128ab#r33405791

cmd/bblfshd/main.go
```diff
@@ -197,9 +197,37 @@ func listenUser(d *daemon.Daemon) {
...
+		return unwrap(err.Err)
+	}
+	return err
+}
+
+func isErrno(err error, eno syscall.Errno) bool {
+	if e, _ := unwrap(err).(syscall.Errno); e == eno {
+		return true
+	}
+	return false
```
bzz: Body of the function can be further simplified to:

```go
e, _ := unwrap(err).(syscall.Errno)
return e == eno
```

dennwc:
Ready for another round

PR #289, opened 2019-05-06
bblfshctl: Add an option to use v2 parsing protocol

### LINKED ISSUE #285: bblfshctl should use protocol v2 for parsing
`bblfshctl parse` [still uses](https://github.com/bblfsh/bblfshd/blob/master/cmd/bblfshctl/cmd/parse.go#L9) protocol v1 for parsing. We must update it to use v2, as all other clients do.

This may be considered a breaking change, thus it should be under `--v2` flag for now.

dennwc:
Add a `--v2` CLI flag to `bblfshctl parse` to use the v2 parsing protocol.

The v1 can only be deprecated with a major version bump, so postpone the change for CLI until #210.

Resolves #285

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -9,6 +9,8 @@ github.com/Nvveen/Gotty v0.0.0-20120604004816-cd527374f1e5/go.mod h1:lmUJ/7eu/Q8
 github.com/alecthomas/template v0.0.0-20160405071501-a0175ee3bccc/go.mod h1:LOuyumcjzFXgccqObfd/Ljyb9UuFJ6TxHnclSeseNhc=
 github.com/alecthomas/units v0.0.0-20151022065526-2efee857e7cf/go.mod h1:ybxpYRFXyAe+OPACYpWeL0wqObRcbAqCMya13uyzqw0=
 github.com/antchfx/xpath v0.0.0-20180922041825-3de91f3991a1/go.mod h1:Yee4kTMuNiPYJ7nSNorELQMr1J33uOpXDMByNYhvtNk=
+github.com/bblfsh/go-client/v4 v4.0.1 h1:1sM/mIchlw872S5jtT4BnPI+oo6eiNaP/B0QedDAchw=
+github.com/bblfsh/go-client/v4 v4.0.1/go.mod h1:Dxot07alCbqwIleIn1AQYiCDfolddYh1z1hhakZhUss=
 github.com/bblfsh/sdk/v3 v3.0.0 h1:bCyTC8gEcMSIo1wzuYpbMYKK9L1KFEzgrEckVvraELc=
 github.com/bblfsh/sdk/v3 v3.0.0/go.mod h1:juMiu8rP3lYJN1e4neEkSyzNieqiFceZzN4AOo0Rm1Q=
 github.com/beorn7/perks v0.0.0-20180321164747-3a771d992973/go.mod h1:Dwedo/Wpr24TaqPxmxbtue+5NUziq4I4S80YR8gNf3Q=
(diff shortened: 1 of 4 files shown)


go.sum
```diff
@@ -232,8 +234,12 @@ golang.org/x/tools v0.0.0-20180221164845-07fd8470d635/go.mod h1:n7NCudcB/nEzxVGm
 golang.org/x/tools v0.0.0-20180917221912-90fa682c2a6e/go.mod h1:n7NCudcB/nEzxVGmLbDWY5pfWTLqBcC2KZ6jyYvM4mQ=
 google.golang.org/genproto v0.0.0-20180716172848-2731d4fa720b h1:mXqBiicV0B+k8wzFNkKeNBRL7LyRV5xG0s+S6ffLb/E=
 google.golang.org/genproto v0.0.0-20180716172848-2731d4fa720b/go.mod h1:JiN7NxoALGmiZfu7CAH4rXhgtRTLTxftemlI0sWmxmc=
+google.golang.org/genproto v0.0.0-20180731170733-daca94659cb5 h1:2PjFmwzH/sxgW9CRJDlEiwMHO8rOk1eMDzVL14HC1e4=
+google.golang.org/genproto v0.0.0-20180731170733-daca94659cb5/go.mod h1:JiN7NxoALGmiZfu7CAH4rXhgtRTLTxftemlI0sWmxmc=
 google.golang.org/grpc v1.13.0 h1:bHIbVsCwmvbArgCJmLdgOdHFXlKqTOVjbibbS19cXHc=
 google.golang.org/grpc v1.13.0/go.mod h1:yo6s7OP7yaDglbqo1J04qKzAhqBH6lvTonzMVmEdcZw=
+google.golang.org/grpc v1.14.0 h1:ArxJuB1NWfPY6r9Gp9gqwplT0Ge7nqv9msgu03lHLmo=
+google.golang.org/grpc v1.14.0/go.mod h1:yo6s7OP7yaDglbqo1J04qKzAhqBH6lvTonzMVmEdcZw=
```
bzz: Interesting, so now it includes multiple version of grpc..

dennwc: Not really. `go.sum` is not a lock file in the usual sense. It records all versions and hashes ever used by the package. So if we decide to revert to gRPC v1.13.0 it will know what hash to expect. And if it's not the same it will error because you won't get the same build as before.

dennwc:
@bzz Well, technically a breaking change. But I doubt anyone is using `bblfshctl` to parse files, so if anyone agrees that it's okay, I'd be happy to make it v2-only.

go.sum
creachadair: I think if we run `go mod tidy` as part of our update procedure, it should clean up hashes that are no longer required to build/test.

dennwc:
OK, then let me remove the v1 part.

Regarding `go mod tidy`, it produced a large diff this time. Probably tried to bump some dependencies. So I intentionally reverted it and will do it separately.

creachadair:
> OK, then let me remove the v1 part.
> 
> Regarding `go mod tidy`, it produced a large diff this time. Probably tried to bump some dependencies. So I intentionally reverted it and will do it separately.

Hmm. In principle `go mod tidy` should not change which versions we depend on, but it does more exhaustively explore the graph, so maybe we were missing some things before. Either way, though I agree it makes sense to consider that separately. I admit I don't fully understand _how_ `tidy` decides which things to revise.

dennwc:
Yep, I've got the feeling that it sometimes tries to re-fetch repositories (`HEAD`? git tags?) and changes versions if it finds that the repository is now compatible with modules.

dennwc:
Done, removed v1 completely.

creachadair:
> Yep, I've got the feeling that it sometimes tries to re-fetch repositories (`HEAD`? git tags?) and changes versions if it finds that the repository is now compatible with modules.

Ah yeah, often that causes the removal of `// indirect` comments, certainly. And if the target does not have semantic version tags, it might have the side-effect of changing the hash too.

creachadair:
It looks like `go mod tidy` also considers build tags, and may expand the footprint for that reason too.

Please don't forget to update the description before merging though. 😄

dennwc:
Sorry, forgot that and now it's too late :(

creachadair:
> Sorry, forgot that and now it's too late :(

It happens. It's not the end of the world. 😉

PR #290, opened 2019-05-06
Update dependencies

dennwc:
Update dependencies and run `go mod tidy`. Follow-up for #289.


creachadair:
> The results look good, but can you please also include replay instructions in the description?
> 
> E.g., was this plain `go mod tidy` or did you also `go get -u`?

I must need a nap or something, it says that right in the original comment, but I somehow overlooked it.

go.mod
```diff
@@ -50,15 +54,20 @@ require (
...
+	go.uber.org/atomic v1.4.0 // indirect
 	golang.org/x/crypto v0.0.0-20190426145343-a29dc8fdc734 // indirect
-	golang.org/x/net v0.0.0-20190424112056-4829fb13d2c6
-	golang.org/x/sys v0.0.0-20190429190828-d89cdac9e872 // indirect
+	golang.org/x/net v0.0.0-20190503192946-f4e77d36d62c
+	golang.org/x/sys v0.0.0-20190506115046-ca7f33d4116e // indirect
 	golang.org/x/text v0.3.2 // indirect
-	google.golang.org/grpc v1.14.0
-	gopkg.in/bblfsh/sdk.v1 v1.16.1
+	google.golang.org/appengine v1.4.0 // indirect
```
bzz: If I'm reading this right - it seems like go now has all the problems of any mature language/build system :)
We are linking in the stuff that will never be used, aren't we? But 🤷‍♂ 

Curious, how long until there is a way for filtering/excluding dependencies.

creachadair: In this case, the indirect dependency is included because one of our dependencies does not have its own module declaration yet. Most of the Go team's /x/ repositories only added support for modules recently—and for x/tools in particular the early versions of their go.mod file had [extraneous dependencies](https://github.com/golang/tools/commit/8dcb7bc8c7fe0a895995c76c721cef79419ac98a). 

So I do not think this is a case of the tool doing something bad. And unlike `dep`, it should not be necessary to filter or exclude dependencies (though you can, with `go mod edit`), but rather the solution to most issues like this will be to fix the error in the dependency and update.

creachadair: Relatedly: I think we should probably take a separate pass of explicitly updating as many of our dependencies as we can, modulo compatibility issues. While doing this for [Kythe](https://gist.github.com/creachadair/28fae4834e3d54022eda7807634838e5), I found most of the weird corner cases went away simply with a more recent tag.

Of course in some cases there are no tags, but I've found that most of the packages that actually get maintained more than once a year are reasonably up-to-date. Tags for the /x/ packages, unfortunately, are stalled on a resolution of https://github.com/golang/go/issues/30241.

dennwc:
@creachadair You were right, it was actually `go get -u` on the SDK and the Go client and `go mod tidy`. At the end, the diff was large because of `-u` part.

PR #292, opened 2019-05-15
Support language protocol v2 and language aliases

### LINKED ISSUE #263: Language aliases support
Enry and Bblfsh use different language names and one driver is responsible for parsing multiple dialects.

Right now bblfsh can accept both bblfsh or enry language names (for most common cases), but returns always only bblfsh ones.

Initial RFC and implementation details are in [mailing list](https://groups.google.com/forum/#!topic/bblfsh-dev/mo1o5d_CwnA)

This issue is going to be used as ☂️ and all details from MailingList will be moved in here before actual work begins (current expectation is `2Q2019`).

## Driver → Language Aliases

 - [bash](https://github.com/bblfsh/bash-driver)
     * shell
     * TODO: add more *sh that enry differenciate
 - [cpp](https://github.com/bblfsh/cpp-driver)
     * C
     * C++
 - [csharp](https://github.com/bblfsh/csharp-driver)
     * C#
 - [Go](https://github.com/bblfsh/go-driver)
 - [Java](https://github.com/bblfsh/java-driver)
 - [JavaScript](https://github.com/bblfsh/javascript-driver)
    * JSX
    * Javascript
 - [PHP](https://github.com/bblfsh/php-driver)
 - [Python](https://github.com/bblfsh/python-driver)
 - [Ruby](https://github.com/bblfsh/ruby-driver)
 - [TypeScript](https://github.com/bblfsh/typescript-driver)

This list covers 2 main cases:
 1. `enry` <=> `bblfshd` language name mismatch (e.g `C#` vs `charp`)
 2. Dialects supported by a single driver (e.g `javascript-driver` for JSX and js, `cpp-driver`  for C\C++, `bash-driver` for *sh)

## Relevant issues
 * initial discussion https://github.com/bblfsh/bash-driver/issues/39
 * [_usage pattern 2_](https://groups.google.com/forum/#!topic/bblfsh-dev/mo1o5d_CwnA) in client https://github.com/src-d/engine/issues/297
 * [_usage pattern 2_](https://groups.google.com/forum/#!topic/bblfsh-dev/mo1o5d_CwnA) in another client https://github.com/src-d/gitbase/issues/728

dennwc:
The change adds support for Language API v2 (see https://github.com/bblfsh/sdk/pull/400) and language aliases.

Language aliases are loaded dynamically when the driver image is first accessed. They stay in memory for the lifetime of `bblfshd` process, as long as no other driver overrides it. Later we should probably return an error if the driver overrides alias of another active driver.

Despite the protocol change, there are no changes in the way how `bblfshd` accesses the list of languages supported by the driver. This is due to the fact that `bblfshd` reads the driver manifest directly from the Docker image and does not use the actual languages API. Still, we want to support Languages API for both `bblfshd` and drivers for symmetry, so clients can access drivers directly when running standalone.

Requires https://github.com/bblfsh/sdk/pull/400

Resolves #263

Unblocks #210

---
This change is [Reviewable](https://reviewable.io/reviews/bblfsh/bblfshd/292)

### DIFF
--- a/daemon/daemon.go
+++ b/daemon/daemon.go
@@ -29,8 +29,9 @@ type Daemon struct {
 	runtime   *runtime.Runtime
 	driverEnv []string
 
-	mu   sync.RWMutex
-	pool map[string]*DriverPool
+	mu      sync.RWMutex
+	pool    map[string]*DriverPool // language ID → driver pool
+	aliases map[string]string      // alias → language ID
 }
 
 // NewDaemon creates a new server based on the runtime with the given version.
@@ -43,6 +44,7 @@ func NewDaemon(version string, build time.Time, r *runtime.Runtime, opts ...grpc
 		build:         build,
 		runtime:       r,
 		pool:          make(map[string]*DriverPool),
+		aliases:       make(map[string]string),
 		UserServer:    grpc.NewServer(opts...),
 		ControlServer: grpc.NewServer(commonOpt...),
 	}
@@ -69,7 +71,9 @@ func registerGRPC(d *Daemon) {
 	protocol1.DefaultService = NewService(d)
 	protocol1.RegisterProtocolServiceServer(d.UserServer, protocol1.NewProtocolServiceServer())
 
-	protocol2.RegisterDriverServer(d.UserServer, NewServiceV2(d))
+	s2 := NewServiceV2(d)
+	protocol2.RegisterDriverServer(d.UserServer, s2)
+	protocol2.RegisterDriverHostServer(d.UserServer, s2)
 	protocol.RegisterService(d.ControlServer, NewControlService(d))
 }
 
@@ -92,7 +96,7 @@ func (d *Daemon) InstallDriver(language string, image string, update bool) error
 		}
 	}
 
-	s, err := d.getDriverImage(context.TODO(), language)
+	s, _, err := d.getDriverImage(context.TODO(), language)
 	if err != nil && !ErrMissingDriver.Is(err) {
 		return ErrRuntime.Wrap(err)
 	}
@@ -118,7 +122,7 @@ func (d *Daemon) InstallDriver(language string, image string, update bool) error
 func (d *Daemon) RemoveDriver(language string) error {
 	driverRemoveCalls.Add(1)
 
-	img, err := d.getDriverImage(context.TODO(), language)
+	img, _, err := d.getDriverImage(context.TODO(), language)
 	if err != nil {
 		return ErrRuntime.Wrap(err)
 	}
@@ -136,6 +140,9 @@ func (d *Daemon) RemoveDriver(language string) error {
 
 func (d *Daemon) DriverPool(ctx context.Context, language string) (*DriverPool, error) {
 	d.mu.RLock()
+	if l, ok := d.aliases[language]; ok {
+		language = l
+	}
 	dp, ok := d.pool[language]
 	d.mu.RUnlock()
 	if ok {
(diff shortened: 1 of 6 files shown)


daemon/daemon.go
```diff
@@ -136,6 +140,9 @@ func (d *Daemon) RemoveDriver(language string) error {
 
 func (d *Daemon) DriverPool(ctx context.Context, language string) (*DriverPool, error) {
 	d.mu.RLock()
+	if l, ok := d.aliases[language]; ok {
+		language = l
+	}
```
bzz: Here and below, as every time we override `language` - we actually use it for routing to an appropriate driver, I think logging it with `DEBUG` log level would also make sense. WDYT?

go.mod
```diff
@@ -69,3 +69,5 @@ require (
 	gopkg.in/bblfsh/sdk.v1 v1.17.0
 	gopkg.in/src-d/go-errors.v1 v1.0.0
 )
+
+replace github.com/bblfsh/sdk/v3 => github.com/dennwc/sdk/v3 v3.0.0-20190515204145-5c71eed53fbf
```
bzz: Nice! Need not to forget to remove it later 😆 

Just curious - was this one added with a text editor, or is it some tooling that was used?

daemon/daemon.go
```diff
@@ -144,40 +151,50 @@ func (d *Daemon) DriverPool(ctx context.Context, language string) (*DriverPool,
...
 	sp, _ := opentracing.StartSpanFromContext(rctx, "bblfshd.runtime.ListDrivers")
 	defer sp.Finish()
 
 	list, err := d.runtime.ListDrivers()
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 
 	for _, d := range list {
```
bzz: As this code looks very structural similar to [`containsLang()`](https://github.com/bblfsh/sdk/pull/400/files#diff-c7921d573f39a395c057e7eeb48f78daR61) - would it be worth to extract a function here as well?

bzz:
Please, help me understand if that intentionally that this work does not touch on usages of [daemon/language.go](https://github.com/bblfsh/bblfshd/blob/master/daemon/language.go) ?

Asking, as I would expect that file not being needed any more, as soon as aliases are in place.

daemon/daemon.go
dennwc: Sounds good, let's do it!

go.mod
dennwc: Sure :) It's nice that it's possible to test CI this way, so I can work on changes to all our project at the same time and then test every part in CI before even finishing the review process. This saves a lot of time.

Can be added with a text editor as:
```
replace github.com/bblfsh/sdk/v3 => github.com/dennwc/sdk/v3 some_branch
```
and then `go mod tidy` will rewrite it to a specific revision. The command may be different since in my case it's done automatically by IDE.

And the second way is to use `go mod` command to add this line:
```
go mod edit -replace github.com/bblfsh/sdk/v3=github.com/dennwc/sdk/v3@some_branch
```

dennwc:
@bzz I considered to remove it, but I think we can leave this code there until we update all the driver with aliases and release at least one `bblfshd` version with those drivers. This way few next `bblfshd` releases will support both old and new drivers in regards to aliases and later we will drop the code you mentioned so only the new drivers will be aliased.

daemon/daemon.go
dennwc: Added this to the write path instead (when the driver is loaded).

go.mod
creachadair: It's worth noting that you can even use `replace` to point to a local file path, if you want. I do this sometimes when I am preparing dependent changes in my personal repositories where I don't want to push the fixes upstream yet.

```
replace github.com/foo/bar => /Users/michael/local/foo/bar
```

creachadair: In re: "Not forgetting to remove patches later":

We could probably add a presubmit check for this, either as a hook or in CI. I'd say it's fair that the committed `go.mod` should not contain any `replace` lines except for work-in-flight. It would be pretty easy to write a tool that flags an error for any such line that isn't accompanied by a comment pointing to an open PR (or some such). Then the CI would protect us. If you think that's worthwhile, I can write it up.

daemon/daemon.go
```diff
@@ -144,40 +151,56 @@ func (d *Daemon) DriverPool(ctx context.Context, language string) (*DriverPool,
 
 	d.mu.Lock()
 	defer d.mu.Unlock()
+	if l, ok := d.aliases[language]; ok {
+		language = l
```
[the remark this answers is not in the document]
dennwc: Yes, Alex already mentioned this and I've added the log to the code path that writes to this map. Don't want to log under a mutex here.

go.mod
dennwc: Sorry, I think I have not communicated my intention well here.

The whole point of me pushing this PR with the `replace` was to test if CI fails with all the changes I've made without waiting for SDK PR to be merged. I'm aware that the line must be removed, and putting the CI check against those `replace` directives would defeat the purpose of doing the PR in the first place. Locally I test with a relative `replace` (`../sdk`), but I assume my environment is not "clean" and there may be issues with the change that CI will catch.

creachadair: > Locally I test with a relative `replace` (`../sdk`), but I assume my environment is not "clean" and there may be issues with the change that CI will catch.

Right—that makes sense. I think I'm the one who confused things: What I was trying to describe is a check that prevents a PR from being _merged_ with the replacement still in place. Putting it in the context of CI wasn't really relevant.

dennwc: Hmm, preventing the merge actually sound interesting, but we'll probably need a separate Github Check for this. Maybe it can be implemented as Lookout analyzer :)

dennwc:
Addressed all comments and switched to a new SDK. Merging when CI is green since the PR is already approved.

bzz:
___
*[daemon/daemon.go, line 143 at r1](https://reviewable.io/reviews/bblfsh/bblfshd/292#-Lf_80v7CDRC1joYkLlY:-Lf_8qBJ13wwpuVbUQgW:bu493fu) ([raw file](https://github.com/bblfsh/bblfshd/blob/53a81f2d959833b8d3305753fd49bc7410230b19/daemon/daemon.go#L143)):*
<details><summary><i>Previously, creachadair (M. J. Fromberger) wrote…</i></summary><blockquote>

Minor: I try to avoid using `l` as a variable name because it's easy to confuse with `I` or `1` in some fonts.
</blockquote></details>

I think it's actually blocking as well!

bzz:
___
*[daemon/daemon.go, line 143 at r1](https://reviewable.io/reviews/bblfsh/bblfshd/292#-Lf_80v7CDRC1joYkLlY:-Lf_9CUM3LP-edsCfr9W:b-qc0lly) ([raw file](https://github.com/bblfsh/bblfshd/blob/53a81f2d959833b8d3305753fd49bc7410230b19/daemon/daemon.go#L143)):*
<details><summary><i>Previously, bzz (Alexander) wrote…</i></summary><blockquote>

I think it's actually blocking as well!
</blockquote></details>

caramba

PR #306, opened 2019-09-12
Kill driver on parse ctx.Done

### LINKED ISSUE #297: Add driver killing capabilities
I have found myself having a problem when trying to parse large amounts of files (~5k files). Basically what happens is that after parsing a large amount of files, the C++ driver stops responding completely, and the Parse request that triggered it, and all subsequent ones, timeout.

After some testing, I've realized that it seems that killing the running driver instance and relaunching it solves the problem, so I would like:

1. This process to be automatically done when there is a user or parse timeout, possibly with a flag when launching the docker image or an environment constant.
2. To be able to kill specific instances by using `bblfshctl instances`, for example by doing `bblfshctl instances --kill ccp` or even `bblfshctl instances --kill-all`

kuba--:
Signed-off-by: kuba-- <kuba@sourced.tech>
This PR partly fixes #297 (on the daemon side). The next step is to try restart a native driver (internally), so it should be done on sdk side.
We may also think about exposing _parse timeout_ flag for bblfshd.

---
This change is [Reviewable](https://reviewable.io/reviews/bblfsh/bblfshd/306)

### DIFF
--- a/daemon/service.go
+++ b/daemon/service.go
@@ -25,6 +25,8 @@ import (
 var (
 	_ protocol2.DriverServer     = (*ServiceV2)(nil)
 	_ protocol2.DriverHostServer = (*ServiceV2)(nil)
+
+	parseKillDelay = time.Second
 )
 
 type ServiceV2 struct {
@@ -72,7 +74,7 @@ func (s *ServiceV2) Parse(rctx xcontext.Context, req *protocol2.ParseRequest) (r
 	req.Language = language
 
 	err = dp.ExecuteCtx(ctx, func(ctx context.Context, driver Driver) error {
-		resp, err = driver.ServiceV2().Parse(ctx, req)
+		resp, err = parseV2(ctx, dp, driver, req)
 		return err
 	})
 	if err != nil {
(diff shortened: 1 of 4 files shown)


daemon/service.go
```diff
@@ -25,6 +25,8 @@ import (
 var (
 	_ protocol2.DriverServer     = (*ServiceV2)(nil)
 	_ protocol2.DriverHostServer = (*ServiceV2)(nil)
+
+	backoffPeriod = time.Second
```
kuba--: shall we rename it? e.g.: `killDelay`?

dennwc: Or `parseKillDelay`

daemon/service.go
```diff
@@ -84,6 +86,36 @@ func (s *ServiceV2) Parse(rctx xcontext.Context, req *protocol2.ParseRequest) (r
...
+	case <-ctx.Done():
+		select {
+		case <-time.After(backoffPeriod):
+			e := ctx.Err()
+			killDriver(pool, drv, e)
+			return nil, e
+
+		case <-done:
+		}
+	case <-done:
```
dennwc: maybe `return` here instead and move the inner select to the bottom

daemon/service.go
```diff
@@ -84,6 +86,36 @@ func (s *ServiceV2) Parse(rctx xcontext.Context, req *protocol2.ParseRequest) (r
 	return resp, err
 }
 
+func parseV2(ctx context.Context, pool *DriverPool, drv Driver, req *protocol2.ParseRequest) (*protocol2.ParseResponse, error) {
+	var (
+		resp *protocol2.ParseResponse
+		err  error
+	)
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
```
dennwc: Do we need `WithCancel` now?

daemon/service.go
```diff
@@ -453,3 +516,9 @@ func (s *ControlService) DriverStates() ([]*protocol.DriverImageState, error) {
 
 	return out, nil
 }
+
+func killDriver(pool *DriverPool, drv Driver, err error) {
```
dennwc: A method on pool? Or maybe an additional parameter to existing `killDriver` method on `pool`?

PR #318, opened 2019-10-23
Adds rootless containers support

ncordon:
Adds rootless containers support and also bumps the version of `libcontainer` from `1.0.0rc5` to current `1.0.0rc9`.

Supersedes #153. It takes over previous @dennwc work. The reason why #153 was failing was due to problems with version rc5 itself, that were solved in [rc6](https://github.com/opencontainers/runc/releases/tag/v1.0.0-rc6) as documented in [libcontainer#1759](https://github.com/opencontainers/runc/pull/1759) and [licontainer#1806](https://github.com/opencontainers/runc/pull/1806).

Note that bumping the `libcontainer` version (it is a problem unrelated to rootless) requires to use `Init: true` for the first process (not sure whether if some process arrives to the container after that with that flag set would cause any damage, probably it would ignored), as documented in [libcontainer#2089](https://github.com/opencontainers/runc/pull/2089) and [libcontainer#1957](https://github.com/opencontainers/runc/issues/1957). Note this change was introduced in rc6, and was not documented until recently (rc8 or even rc9 :roll_eyes:). I had to actually navigate and debug the `libcontainer` code to look why the containers were not starting correctly, and later I searched for known issues related to `Init` flag in the bug tracker of the project.

Also, this may require `sudo` access in some OSs to enable unprivileged containers support (not Ubuntu for example), as documented in the [buildah documentation](https://wiki.archlinux.org/index.php/Buildah#Enable_support_to_build_unprivileged_containers) or in the [usernetes documentation](https://github.com/rootless-containers/usernetes#distribution-specific-hint). It may we worth linking this information in the `README` if we proceed forward with this PR or a hint of what it may be needed to enable such unprivileged containers:

```sh
# For only the current session
sudo sysctl kernel.unprivileged_userns_clone=1
```

```sh
# For a permanent change
echo "kernel.unprivileged_userns_clone=1" >> /etc/sysctl.conf
sudo sysctl -p
```

With all that being said, as documented in the [Docker docs](https://docs.docker.com/engine/security/seccomp/), the security policy by default disables `unshare` command inside the containers (which is needed to create a user namespace and also to give a `Hostname` to the container as a searchable driver). Therefore need to run `bblfshd` with `--security-opt seccomp=unconfined`. Also, as documented in [libcontainer#1658](https://github.com/opencontainers/runc/issues/1658), there is a problem with spawning rootless containers inside another non-root container and the `/proc` mount / masking. Adding a volume `-v /proc:/newproc` would solve that problem. I am going to try to come up with a nicer default config to squash the aforementioned two bugs. Right now this works:

```sh
# Builds bblfsh/bblfshd:dev-924c14e
make build

docker run \
  --name bblfshd -p 9432:9432 \
  -v /var/lib/bblfshd:/var/lib/bblfshd \
  -v /proc:/newproc \
  --security-opt seccomp=unconfined \
  bblfsh/bblfshd:dev-924c14e
```
```sh
# Managing drivers works
docker exec -it bblfshd bblfshctl driver remove --all
docker exec -it bblfshd bblfshctl driver install --recommended
# Parsing from inside the container works
docker exec -it bblfshd bblfshctl parse /opt/bblfsh/etc/examples/python.py
# Parsing from outside the container works
curl https://raw.githubusercontent.com/bblfsh/sdk/master/uast/types.go -o types.go
bblfsh-cli types.go
```

---
This change is [Reviewable](https://reviewable.io/reviews/bblfsh/bblfshd/318)

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,25 +18,56 @@ This project is now part of [source{d} Engine](https://sourced.tech/engine),
 which provides the simplest way to get started with a single command.
 Visit [sourced.tech/engine](https://sourced.tech/engine) for more information.
 
+#### Rootless mode
+
 The recommended way to run *bblfshd* by itself is using Docker:
 
 ```sh
-docker run -d --name bblfshd --privileged -p 9432:9432 -v /var/lib/bblfshd:/var/lib/bblfshd bblfsh/bblfshd
+docker run -d --name bblfshd \
+  -p 9432:9432 \
+  -v /var/lib/bblfshd:/var/lib/bblfshd \
+  -v /proc:/newproc \
+  --security-opt seccomp=./bblfshd-seccomp.json \
+  bblfsh/bblfshd
 ```
 
 On macOS, use this command instead to use a Docker volume:
 
 ```sh
-docker run -d --name bblfshd --privileged -p 9432:9432 -v bblfsh-storage:/var/lib/bblfshd bblfsh/bblfshd
+docker run -d --name bblfshd \
+  -p 9432:9432 \
+  -v bblfsh-storage:/var/lib/bblfshd bblfsh/bblfshd \
+  -v /proc:/newproc \
+  --security-opt seccomp=./bblfshd-seccomp.json \
+  bblfsh/bblfshd
 ```
 
 
-The container should be executed with the `--privileged` flag since *bblfshd* is
-based on [container technology](https://github.com/opencontainers/runc/tree/master/libcontainer)
-and interacts with the kernel at a low level. *bblfshd* exposes a gRPC server at
-the port `9432` by default which is used by the [clients](https://github.com/search?q=topic%3Aclient+org%3Abblfsh&type=Repositories)
+To understand the flags `-v /proc:/newproc` and `--security-opt seccomp=./bblfshd-seccomp.json`, 
+where [`bblfshd-seccomp.json`](./bblfshd-seccomp.json) is a file present in this repo, and check 
+further requirements, please refer to [rootless.md](./rootless.md). `bblfshd` is based on 
+[container technology](https://github.com/opencontainers/runc/tree/master/libcontainer)
+and interacts with the kernel at a low level. It exposes a gRPC server at the port `9432` by default 
+which is used by the [clients](https://github.com/search?q=topic%3Aclient+org%3Abblfsh&type=Repositories)
 to interact with the server. Also, we mount the path `/var/lib/bblfshd/` where
 all the driver images and container instances will be stored.
+
+#### Privileged mode
+
+We advise against it, but if you prefer to run `bblfshd` in `privileged` mode to skip configuration steps of 
+[rootless.md](rootless.md), you could do, in Linux:
+
+```sh
+docker run -d --name bblfshd --privileged -p 9432:9432 -v /var/lib/bblfshd:/var/lib/bblfshd bblfsh/bblfshd
+```
+
+or macOs:
+
+```sh
+docker run -d --name bblfshd --privileged -p 9432:9432 -v bblfsh-storage:/var/lib/bblfshd bblfsh/bblfshd
+```
+
+#### Install drivers
 
 Now you need to install the driver images into the daemon, you can install
 the official images just running the command:
--- a/bblfshd-seccomp.json
+++ b/bblfshd-seccomp.json
@@ -1,0 +1,804 @@
+{
+	"defaultAction": "SCMP_ACT_ERRNO",
+	"archMap": [
+		{
+			"architecture": "SCMP_ARCH_X86_64",
+			"subArchitectures": [
+				"SCMP_ARCH_X86",
+				"SCMP_ARCH_X32"
+			]
+		},
+		{
+			"architecture": "SCMP_ARCH_AARCH64",
+			"subArchitectures": [
+				"SCMP_ARCH_ARM"
+			]
+		},
+		{
+			"architecture": "SCMP_ARCH_MIPS64",
+			"subArchitectures": [
+				"SCMP_ARCH_MIPS",
+				"SCMP_ARCH_MIPS64N32"
+			]
+		},
+		{
+			"architecture": "SCMP_ARCH_MIPS64N32",
+			"subArchitectures": [
+				"SCMP_ARCH_MIPS",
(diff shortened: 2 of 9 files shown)


ncordon:
Special ping to @dennwc (since I completed his work) to test this branch and suggest possible changes or discard this completely

ncordon:
___
*[runtime/runtime.go, line 160 at r3](https://reviewable.io/reviews/bblfsh/bblfshd/318#-LrsT-cR7_RnuJzx4Wnn:-Lrt6zP60sq8q_vnzdtV:bq204um) ([raw file](https://github.com/bblfsh/bblfshd/blob/bc35c1d62051f05c45e2105675e554c64371eddd/runtime/runtime.go#L160)):*
<details><summary><i>Previously, dennwc (Denys Smirnov) wrote…</i></summary><blockquote>

Can you please verify that all existing drivers work with this change?
</blockquote></details>

They seem to be working, yes (at least `bblfsh-cli` can parse files for each language we support in `recommended` drivers). Do you suggest testing anything else?

ncordon:
*[a discussion](https://reviewable.io/reviews/bblfsh/bblfshd/318#-LrsTpqg94w66S0Km39C:-LryAX4Y8Z_Q3qjvX8cf:b-896fja) (no related file):*
<details><summary><i>Previously, dennwc (Denys Smirnov) wrote…</i></summary><blockquote>

Let's update the `README` as well with all the things you mentioned. Maybe adding a `rootless.md` and linking to it will be a good idea.
</blockquote></details>

Done!

---

ncordon:
___
*[daemon/driver.go, line 77 at r3](https://reviewable.io/reviews/bblfsh/bblfshd/318#-LrtN-lh2jHMA6O98YOS:-LryAbFN7g07FGaBLFV5:b-896fja) ([raw file](https://github.com/bblfsh/bblfshd/blob/924c14eb9e31bffb0162759fab588982470447e0/daemon/driver.go#L77)):*
<details><summary><i>Previously, creachadair (M. J. Fromberger) wrote…</i></summary><blockquote>

Some of the files in this PR will need a gofmt before merging.
</blockquote></details>

Done!

ncordon:
___
*[runtime/container.go, line 89 at r4](https://reviewable.io/reviews/bblfsh/bblfshd/318#-LryAfGq5ve6rbdX4ZIl:-LryAfGq5ve6rbdX4ZIm:b-h6u2q7) ([raw file](https://github.com/bblfsh/bblfshd/blob/6740fc8e5045489a19ad2691345a15feb03a0b42/runtime/container.go#L89)):*
> ```Go
> 
> func (c *container) Stop() error {
> 	if err := c.process.Signal(syscall.SIGKILL); err != nil {
> ```

If I use `SIGTERM` instead of `SIGKILL` here, we see errors in `bblfshd` logs when running without `--privileged` flags about drivers not being stopped. In other words, in rootless containers the containers processes I spin ignore the `SIGTERM` signal.

ncordon:
___
*[runtime/container.go, line 89 at r4](https://reviewable.io/reviews/bblfsh/bblfshd/318#-LryAfGq5ve6rbdX4ZIl:-LryGP8eEXOO6g_BveuA:b-896fja) ([raw file](https://github.com/bblfsh/bblfshd/blob/6740fc8e5045489a19ad2691345a15feb03a0b42/runtime/container.go#L89)):*
<details><summary><i>Previously, dennwc (Denys Smirnov) wrote…</i></summary><blockquote>

Killing sounds like a good idea in general. We don't care about any state in the driver's memory.

Can you please add a comment, so no one reverts it by accident?
</blockquote></details>

Done!

ncordon:
___
*[bblfshd-seccomp.json, line 55 at r5](https://reviewable.io/reviews/bblfsh/bblfshd/318#-LryD1KuAgYEFwDlr1Qi:-LryGU_XC_u29tHWGrWC:b-dl70ic) ([raw file](https://github.com/bblfsh/bblfshd/blob/643340619e5e905536402dc1b18a865bbf785708/bblfshd-seccomp.json#L55)):*
<details><summary><i>Previously, dennwc (Denys Smirnov) wrote…</i></summary><blockquote>

Hmm, interesting. We should review this and narrow down the list (in future PRs).
</blockquote></details>

Nothing against it of course :ok_hand: . Just to clarify, this syscalls are the ones docker is allowed by default ([default.json](https://github.com/moby/moby/blob/master/profiles/seccomp/default.json)) and are considered secure, plus `mount, unshare, pivot_root, keyctl, umount2`and `sethostname` (those last ones are needed for sure because I tested removing any of them an we cannot spawn containers inside anymore. It is better than using `seccomp=unconfined` anyway, which does not restrict syscalls. And using `--privileged` is even less restrictive than `seccomp=unconfined` because you can do a lot of dangerous stuff with the host inside the containers.

ncordon:
___
*[README.md, line 40 at r6](https://reviewable.io/reviews/bblfsh/bblfshd/318#-LrzzcqEEPldvBrtw7uh:-LrzzcqFApz8berdQggu:bccx9m4) ([raw file](https://github.com/bblfsh/bblfshd/blob/6dd10246c56fea4f8b68b85a48243841a3baf1b3/README.md#L40)):*
> ```Markdown
>   -p 9432:9432 \
>   -v bblfsh-storage:/var/lib/bblfshd bblfsh/bblfshd \
>   -v /proc:/newproc \
> ```

I am realizing, would this line `-v/proc:/newproc` work under macOS? @kuba-- @creachadair

ncordon:
___
*[README.md, line 40 at r6](https://reviewable.io/reviews/bblfsh/bblfshd/318#-LrzzcqEEPldvBrtw7uh:-Ls1NerY3TQHF8XZVNta:b-2ussgq) ([raw file](https://github.com/bblfsh/bblfshd/blob/6dd10246c56fea4f8b68b85a48243841a3baf1b3/README.md#L40)):*
<details><summary><i>Previously, ncordon (Nacho Cordón) wrote…</i></summary><blockquote>

So can we suppose it is safe to omit that part? I think we do not have way of testing it, since I do not have a macOs and people with macOS cannot do `make build` and test the new container :roll_eyes: 
</blockquote></details>

I think it is safe to leave it there, because the folder would be looked for inside the Moby VM that runs docker on macOS, according to this: https://docs.docker.com/docker-for-mac/osxfs/#namespaces

PR #324, opened 2019-10-24
Print content hashes in logs

### LINKED ISSUE #302: Log file hash when parsing files
When parsing a file, it will be nice to log the content hash.

This will help us find specific files that fail during large parse jobs.

dennwc:
Tracking down bugs in driver is hard if all we get in logs is the filename. To help with it, also log the file content.

The proposal is to log both SHA1 and Git hash. The reason is outlined in https://github.com/bblfsh/bblfshd/issues/302, but in short, we want to be able to search Git itself for the file using the Git hash, and also see the original file hash (as reported by `sha1sum`) in case files are not originating from Git (e.g. sent from CLI or the client).

The open question is, do we want SHA1 specifically (apart from a Git hash)?

MD5 might be much faster, but may result in collisions. At the same time, it's returned as a metadata field in cloud storage services provided by Google and AWS.

If not MD5, Blake2 may be a good alternative both in terms of performance and avoiding collisions. The downside is, of course, that it's not used as widely as SHA1.

Resolves #302

---
This change is [Reviewable](https://reviewable.io/reviews/bblfsh/bblfshd/324)

### DIFF
--- a/daemon/service.go
+++ b/daemon/service.go
@@ -4,8 +4,12 @@ package daemon
 
 import (
 	"context"
+	"crypto/sha1"
+	"encoding/hex"
 	"errors"
 	"fmt"
+	"io"
+	"strconv"
 	"time"
 	"unicode/utf8"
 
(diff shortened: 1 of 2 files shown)


ncordon:
:lgtm:

Issue #301, opened 2019-07-21
Read the logging level from an environment var


dpordomingo:
caused by https://github.com/src-d/sourced-ce/issues/142

`bblfsh` logging level is only configurable through command arguments, so when directly using the docker image, configuring the logging level is not that simple as changing an env value from the outside.

creachadair:
This seems like a reasonable request, but I would rather not have to re-implement this from first principles. What do other teams do, to support this use case?

dpordomingo:
`gitbase` uses [`go-flags`](https://github.com/jessevdk/go-flags), as suggested by [our guides, about CLI tools](https://github.com/src-d/guide/blob/master/engineering/conventions/go.md#cli)

But newer `gitcollector`[[1]](https://github.com/src-d/gitcollector/blob/master/cmd/gitcollector/main.go#L13), `ghsync`[[2]](https://github.com/src-d/ghsync/blob/master/cmd/ghsync/main.go#L17) or `borges`[[3]](https://github.com/src-d/borges/blob/master/cli/borges/main.go#L17) use [`go-cli`](https://github.com/src-d/go-cli), that already wraps `go-flags` but also _reduces boilerplate code and helps in being more homogeneous with respect how our CLI work and look like_.

creachadair:
@dpordomingo Thanks for the pointers! It looks like the important contribution from `go-cli` is bringing in `go-log`. @kuba-- has already started some work on migrating us over to use `go-log`, so hopefully we will be able to get environment support more easily.

dpordomingo:
How is the migration going?

Would it be a simple alternative something like this [bblfsh/web#227](https://github.com/bblfsh/web/pull/227/files#diff-157b0e016d598e869df099dff16c041bR44)?
In summary, the logging level could be read from the environment and set if no `log-level` arg was passed at [cmd/bblfshd/main.go#L84](https://github.com/bblfsh/bblfshd/blob/master/cmd/bblfshd/main.go#L84)

Doing so, it would also unblock https://github.com/src-d/backlog/issues/1485

creachadair:
> Would it be a simple alternative something like this [bblfsh/web#227](https://github.com/bblfsh/web/pull/227/files#diff-157b0e016d598e869df099dff16c041bR44)?
[...]

PR #327, opened 2019-10-28
Read default log level from the environment.

creachadair:
This is intended to address part of #301, as requested by @dpordomingo.

Note that this only affects the daemon, _not_ the drivers.

---
This change is [Reviewable](https://reviewable.io/reviews/bblfsh/bblfshd/327)


creachadair:
The coverage failure does not appear related to this change; the log complains about changes in daemon/pool.go. My inclination would be to merge it as-is given that the rest of the CI passed.

dpordomingo:
great!!
do you have any plan to release this in the short term?

dennwc:
Yes, I think we can make a patch release, if necessary.

